PR #12, opened 2019-07-21
Add the teacher's "large display" view

munnsmunns:


src/app/large-display/large-display.component.html
```diff
@@ -0,0 +1,4 @@
+<div id="largeDisplayContent">
+    <img id="prairieHill" src="/assets/images/Backgrounds/prairieHill.jpg">
```
floogulinc: `backgrounds` should probably be lowercase.

angular.json
```diff
@@ -27,6 +27,11 @@
               "src/assets",
               "src/manifest.webmanifest"
             ],
+            "stylePreprocessorOptions": {
+              "includePaths": [
+                    "node_modules"
+               ]
+            },
```
floogulinc: This is not needed. It seems to work fine without it and we should probably avoid including everything in `node_modules`.

src/app/timer-bar/timer-bar.component.scss
```diff
@@ -0,0 +1,37 @@
...
+
+$bufferColor: #A1A1A1;
+
+mdc-linear-progress {
+    height: 40px;
+    border-radius: 16px;
+}
+
+// The only way I've found to change the buffer color on the progress bar
+::ng-deep .mdc-linear-progress__buffer {
```
floogulinc: `::ng-deep` is [deprecated](https://angular.io/guide/deprecations#deep-component-style-selector)

src/styles.scss
```diff
@@ -12,4 +12,11 @@ $mdc-theme-secondary: #FFC107; // secondary color
         left: none !important;
     }
 
+}
+
+// Changes the buffer color to a darker color so the text is more readable
+// Note: I tried to do this in the component scss file where it would make more sense to do so, but it did not work for some reason.
```
floogulinc: I'm looking into structuring our SCSS to allow for global variables and mixins.

munnsmunns:
Go to localhost:4200/teacher to view the large display stuff if you want to check it out

floogulinc:
In general you should probably be using classes instead of IDs for styling.

src/app/large-display/large-display.component.ts
```diff
@@ -8,13 +8,21 @@ import { TimerBarComponent } from 'src/app/timer-bar/timer-bar.component'
...
+
   @ViewChild('timerBar', {static: true}) timerBar: TimerBarComponent;
-  constructor() {   
-    
+
+  constructor() { }
+
+  // resize the component height with the window
+  @HostListener('window:resize', ['$event']) onResize(event) {
+    this.componentHeight = window.innerHeight-83;
```
floogulinc: I feel like you could do the resizing with CSS instead of this.

floogulinc:
Won;t the actual timer incrementing and stuff be handled by an external controller? I don't think that logic needs to be part of the component.

src/app/large-display/large-display.component.ts
munnsmunns: Setting it to 100% was the first thing I tried when I started working on it yesterday. Changing the height in the css would be preferable, but it doesn't work the way you would expect.

src/app/fullscreen-button/fullscreen-button.component.ts
```diff
@@ -0,0 +1,61 @@
...
+// Code from this stackoverflow issue: https://stackoverflow.com/questions/51998594/how-to-make-google-chrome-go-full-screen-in-angular-4-application
+@Component({
+  selector: 'app-fullscreen-button',
+  templateUrl: './fullscreen-button.component.html',
+  styleUrls: ['./fullscreen-button.component.scss']
+})
+export class FullscreenButtonComponent implements OnInit {
+
+  elem;
+  fullscreen: boolean;
```
floogulinc: We should probably check for if the browser is fullscreen instead of keeping an internal variable for it. The user can enter fullscreen in many different ways (F11 for example).

michael-small:
Screens as big as iPads crunch up the months in the time bar over each other.

munnsmunns:
It could probably be modified so that on thinner screens the text is a smaller size. If we do want to support smaller screens that is

floogulinc:
Where does the background image come from? Do we have the right to use it?

floogulinc:
We should probably at least assume that this will be on displays ranging from 1024x768 to 1920x1080.

src/app/large-display/large-display.component.ts
```diff
@@ -0,0 +1,28 @@
...
+@Component({
+  selector: 'app-large-display',
+  templateUrl: './large-display.component.html',
+  styleUrls: ['./large-display.component.scss']
+})
+export class LargeDisplayComponent implements OnInit {
+
+  componentHeight: number;
+
+  @ViewChild('timerBar', {static: true}) timerBar: TimerBarComponent;
```
floogulinc: You should generally avoid accessing the methods of a component this way. The timer bar component shouldn't handle the timing itself anyway and should be passed in a time to show as an input.

PR #21, opened 2020-03-26
Generate interfaces for integrating Fire base 

wallerli:
Currently the interfaces of Bee, Flower, and Nest are created. Another Item interface is only used for displaying things on the large display for now.

This PR is only for interfaces to be generated for fire base integration. See #22 for modifications to the large display.


src/app/flower.ts
```diff
@@ -0,0 +1,9 @@
+export interface FLower {
```
floogulinc: should be `Flower`

src/app/flower.ts
```diff
@@ -0,0 +1,9 @@
...
+  _id: string;
```
floogulinc: `id` doesn't need an underscore

PR #22, opened 2020-03-26
Large Display and some review screens

wallerli:
Currently the timer bar component controls the game progress. The large display uses its returned values to display things on the screen and pass values to the progress indicator to show the progress.

Modifications are needed after the fire base is there.


wallerli:
About the current component interaction in teacher's side: For now the timer bar is the actual component that controls the game time. The large display only reads the time from the timer bar component and the progress indicator component import the time from the large display.

wallerli:
Currently, when attempting a rewinding, the timer bar component is requiring a browser confirmation to make the new game time public to other components or reset to a time before rewinding.
The large display is now using a different way to store flowers / bees, we can switch it to use the interfaces that we implemented after that is 'finalized'.

src/app/display-item/display-item.component.html
```diff
@@ -0,0 +1,2 @@
+<img class="active_{{this.species.active}}" [@active]="species.active" [@position]="{value: species.displayState, params: {left: species.x, top: species.y, scale: species.scale, offset: species.offset }}"
```
floogulinc, line 1: We should try to use `NgClass` for this.

PR #30, opened 2020-05-22
Add GH Actions workflow for unit tests

floogulinc:


floogulinc:
We won't be testing everything but there is a lot of modular logic stuff that we can do unit testing for.

helloworld12321:
Am I remembering correctly that we're waiting to update the tests until after the Timer refactoring has gone through?

floogulinc:
There are probably many components we simply won't test (like the AR one) but we have some tests for our new timer stuff so it would be nice to have the GH Actions workflow for that.

floogulinc:
Once this is merged I will be pulling it into the timer branch so we can have it running those tests.

PR #33, opened 2020-05-27
Add "Update Angular" GH Actions workflow

floogulinc:
This will run weekly on Monday at 12:00 UTC. It runs `ng update` and makes a pull request if there are any updates for Angular libraries available.


floogulinc:
I fixed a small thing @helloworld12321, can you review again?

PR #48, opened 2020-06-17
Move the timer progress bar to the bottom

helloworld12321:
At Paul's request, I've made this change in a side-branch so we can decide if we like it or not :)


helloworld12321:
Okay, the verdict is in—we do indeed want the bar at the bottom :)

PR #52, opened 2020-06-18
Set up firebase

### LINKED ISSUE #51: Set up Firebase
Import the right packages and configure all of the configuration :upside_down_face:

helloworld12321:
Closes #51.


helloworld12321:
(As part of installing AngularFire, we need to upgrade to Angular version 9.)

Issue #54, opened 2020-06-23
Log in anonymously as a student


helloworld12321:

kklamberty:
Not yet explored: 
- [ ] Naming an anonymous user (setting a displayName)
- [ ] Put the user into the database, probably by session

PR #60, opened 2020-06-23
Create authentication service and test page for session

kklamberty:
We generated the component for a test session page and generated an auth service.

Works on issue #54.

Co-authored-by: Audrey Le Meur <lemeu001@umn.edu>
Co-authored-by: Erik Rauer <rauer007@morris.umn.edu>


src/app/auth.service.ts
```diff
@@ -0,0 +1,27 @@
+import { Injectable } from '@angular/core';
+import { AngularFireAuth } from '@angular/fire/auth';
+import { FirebaseService } from './firebase.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class AuthService {
+
+  userCredential: Promise<void | firebase.auth.UserCredential>;
```
floogulinc, line 10: We probably don't need to keep track of this
```suggestion
```

src/app/auth.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  constructor(public auth: AngularFireAuth, public firebaseService: FirebaseService) { }
```
floogulinc, line 12: These shouldn't be public
```suggestion
  constructor(private auth: AngularFireAuth, private firebaseService: FirebaseService) { }
```

src/app/auth.service.ts
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * Observable that emits the currently logged in firebase.User. Is undefined if no one is logged in.
+   */
+  getCurrentUser$ = this.auth.authState;
+
+  /**
+   * Checks to see if the user is already logged in. If they aren't, it will create a new anonymous account.
+   */
+  logStudentIn() {
+    this.userCredential =  this.auth.auth.signInAnonymously().catch((error) => {
```
floogulinc, line 23: We should probably return this and not store it in the service.
```suggestion
    return this.auth.auth.signInAnonymously().catch((error) => {
```

src/app/session-page-test/session-page-test.component.ts
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * Calls firebase service to add currently logged in user and their preferred name to the database
+   * @param name Student's name
+   */
+  addStudentToDatabase(name) {
+    this.authService.getCurrentUser$.subscribe( user => {
+      this.userID = user.uid;
+      this.firebaseService.addStudent(this.userID, {name});
+    });
+  }
```
floogulinc, line 29: This probably should be a method on the auth service

src/app/firebase.service.ts
```diff
@@ -1,15 +1,26 @@
...
   allSessions$: Observable<any[]> = this.firestore.collection('sessions').valueChanges();
 
+  /**
+   * Adds student to firestore
+   * @param id Student' id
+   * @param studentInfo map of student's information including name
+   */
+  addStudent(id: string, studentInfo: { name?: string}) {
+    this.firestore.collection('sessions/' + this.currentSession + '/students').doc(id).set(studentInfo);
+  }
```
floogulinc, line 25: Consider naming this `addStudentToSession` and taking a session ID as a parameter.

PR #63, opened 2020-06-26
Let the student read round data from Firebase

helloworld12321:


src/app/firebase.service.ts
```diff
@@ -1,13 +1,43 @@
...
+      .doc<Session>(id)
+      .snapshotChanges()
+      .pipe(map(action => ({id: action.payload.id, ...action.payload.data()})));
+  }
+
+  /**
+   * Return an observable stream of the round data for the round whose
+   * Firebase ID is `roundId` within the session denoted by `sessionId`.
+   */
+  public getRound(roundPath: {sessionId: string, roundId: string}): Observable<FirebaseRound> {
```
floogulinc, line 35: We should probably make this just take `sessionId` and `roundId` as separate parameters, if we have them as an object we can always use `...roundPath`.

helloworld12321: You know, I tried that first, but the `...roundPath` doesn't work—JavaScript isn't smart enough to associate object keys with parameter names :-(

See also: https://stackoverflow.com/questions/48008539/how-to-spread-an-object-to-a-function-as-arguments

floogulinc: Ah ok. Nevermind then.

PR #67, opened 2020-07-01
Login page

audreylemeur:


src/app/student-login/student-login.component.scss
```diff
@@ -0,0 +1,28 @@
...
+    background-size: cover;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+}
+
+.login-form{
+  min-width: 40%;
+  max-width: 60%;
+  max-height: 50%;
```
helloworld12321, line 18:
```suggestion
```
Consider possibly removing this rule—hard-capping the size of the form to 50% won't resize the card inside. (The card will just spill over outside of the form. :man_shrugging:)

floogulinc:
I would still like to review this before it's merged btw

src/app/student-login/student-login.component.scss
helloworld12321: (And if that happens, the alignment gets a bit wonky:
image
)

src/app/home/home.component.html
```diff
@@ -9,5 +9,6 @@
     <li><a routerLink="/firebase-test">firebase-test</a></li>
     <li><a routerLink="/session-page-test">session-page-test</a></li>
     <li><a routerLink="/json-data-test">json-data-test</a></li>
+    <li><a routerLink="/join-session">join-session</a></li>
```
floogulinc, line 12:
```suggestion
    <li><a routerLink="/join">join-session</a></li>
```

src/app/app-routing.module.ts
```diff
@@ -17,7 +18,8 @@ const routes: Routes = [
   {path: 'flower-test', component: FlowerTestComponent},
   {path: 'firebase-test', component: FirebaseTestComponent},
   {path: 'session-page-test', component: SessionPageTestComponent},
-  {path: 'json-data-test', component: JsonDataTestComponent}
+  {path: 'json-data-test', component: JsonDataTestComponent},
+  {path: 'join-session', component: StudentLoginComponent}
```
floogulinc, line 22:
```suggestion
  {path: 'join', component: StudentLoginComponent}
```

src/app/student-login/student-login.component.html
```diff
@@ -0,0 +1,22 @@
...
+      <mat-card-title>Join Session</mat-card-title>
+      <mat-card-content class="login-card-content">
+        <mat-form-field appearance="outline" class='form-field'>
+          <mat-label>Name</mat-label>
+          <input matInput class='name'  formControlName="nameControl">
+        </mat-form-field>
+        <mat-form-field appearance="outline" class='form-field'>
+          <mat-label>Session ID (Not implemented yet)</mat-label>
+          <input matInput class='session' formControlName="sessionControl">
+        </mat-form-field>
```
floogulinc, line 13: I think we should have the session ID first and the Name field second

src/app/student-login/student-login.component.scss
```diff
@@ -0,0 +1,28 @@
+@import "../../variables";
+
+
+.page{
```
floogulinc, line 4: We should avoid too generic classes like `.page`

src/app/student-login/student-login.component.scss
floogulinc, line 18:
```suggestion
  width: 400px;
  margin: 16px;
```

The `width` here is essentially a maximum width for the card, it still scales down on narrower screens but also doesn't get super wide on larger screens. The margin gives a margin around it for smaller screens. See this comparison on an iPhone SE:

| Current | Suggested |
|--|--|
| buzz-about-git-login-page mn-pollinators vercel app_join-session(iPhone 5_SE) (1) | localhost_4200_join-session(iPhone 5_SE) |

src/app/student-login/student-login.component.scss
```diff
@@ -0,0 +1,28 @@
...
+
+.login-form{
+  min-width: 40%;
+  max-width: 60%;
+  max-height: 50%;
+}
+
+.login-card{
+  padding: 20px;
+}
```
floogulinc, line 23:
```suggestion
```

This seems unnecessary.

src/app/student-login/student-login.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor(public authService: AuthService) {
+    this.sessionID = 'demo-session'; // Temporary until multiple sessions are supported
+  }
+
+  ngOnInit(): void {
+    this.authService.logStudentIn();
+    this.nameControl = new FormControl();
+    this.sessionControl = new FormControl();
+    this.sessionFormGroup = new FormGroup({nameControl: this.nameControl, sessionControl: this.sessionControl});
```
floogulinc, line 26: This can be done directly when they are declared rather than in `ngOnInit`.

floogulinc: Also since you aren't using the individual `FormControl`s anywhere you can just create them right in the constructor for the `FormGroup`. Something like this:
```ts
sessionFormGroup: FormGroup = new FormGroup({
    nameControl: new FormControl(),
    sessionControl: new FormControl(),
  });
```

We should also add some validation to these form inputs. At the very least making both inputs required to click "Join".

src/app/student-login/student-login.component.scss
```diff
@@ -0,0 +1,28 @@
...
+.page{
+    background-repeat: no-repeat;
+    background-image: url('/assets/images/large-display/hill-compressed.jpg');
+    height: 100%;
+    width: 100%;
+    background-size: cover;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+}
```
floogulinc, line 13: Since we will probably be using this type of background in many places we should move the parts of this that set the background image and position it to a new class in the global styles.

src/app/student-login/student-login.component.scss
helloworld12321: Ooh, I like the look of that!

src/app/student-login/student-login.component.scss
ErikRauer: Do we also want to be setting the height and width in this class? At the moment I've just moved the three background styles, but it doesn't show the full image when applied to elements that are way wider than they are tall or vice versa (see [the session test page](https://buzz-about-gg557td96.vercel.app/session-page-test))

floogulinc: Yeah I think the height and width should be left up to the individual component. The way you did it seems right. One if the ideas of CSS is that you can use multiple classes like this where each one does a small thing.

PR #69, opened 2020-07-06
Let the teacher populate the round with data

### LINKED ISSUE #36: Synchronize timer between teacher and students

### LINKED ISSUE #49: Room controller service

helloworld12321:
We're not actually dealing with _creating_ rounds or sessions at the moment—just setting data (ie, the list of flowers, the time) on a pre-existing round.

Closes #49.
Closes #36.


src/app/large-display/large-display.component.ts
```diff
@@ -111,31 +151,82 @@ export class LargeDisplayComponent implements OnInit, AfterViewInit {
...
 
+    // Make sure that when the timer ticks, it updates the round in Firestore.
+    this.timerService.running$.subscribe(running => {
+      this.firebaseService.updateRoundData(this.roundPath, {running});
+    });
+    this.timerService.currentTime$.subscribe(timePeriod => {
+      this.firebaseService.updateRoundData(this.roundPath, {
+        currentTime: timePeriod.time,
+      });
+    });
```
floogulinc, line 229: We will definitely need to be careful about unsubscribing these

src/app/large-display/large-display.component.ts
```diff
@@ -111,31 +151,82 @@ export class LargeDisplayComponent implements OnInit, AfterViewInit {
...
       tickSpeed: 1000,
       currentTime: this.startTime,
       endTime: this.endTime
     });
 
+    // Keep a copy of the "running" boolean for the large display controls to
+    // use.
     this.timerService.running$.subscribe(running => {
       this.running = running;
     });
```
floogulinc, line 219: I feel like if things in the large display can use the `async` pipe, they should. If we need subscriptions like this we should make sure they get unsubscribed properly

src/app/large-display/large-display.component.ts
helloworld12321: Very much agreed—we now have tests to make sure that we're not writing to the database after the component is destroyed.

helloworld12321: (The subscriptions themselves have been refactored so that they only fire if we're in a round; otherwise, they just stay dormant.)

src/app/large-display/large-display.component.ts
helloworld12321: Done, thanks!

src/app/firebase.service.ts
```diff
@@ -75,4 +75,21 @@ export class FirebaseService {
...
+   * (That way you can never end up with a half-initialized round document;
+   * See https://github.com/angular/angularfire/pull/1247#issuecomment-336226671)
+   *
+   * @param path The Firestore ID of the round to modify, along with with the
+   *   ID of the session that it lives in.
+   * @param data The new round data.
+   */
+  updateRoundData(id: RoundPath, data: Partial<FirebaseRound>) {
+    this.getRoundDocument(id).update(data);
+  }
```
floogulinc, line 94: The parameter `id` should be `roundPath` to be consistent with the other functions in this service.
```suggestion
  updateRoundData(roundPath: RoundPath, data: Partial<FirebaseRound>) {
    this.getRoundDocument(roundPath).update(data);
  }
```

src/app/large-display/large-display.component.ts
```diff
@@ -111,34 +153,107 @@ export class LargeDisplayComponent implements OnInit, AfterViewInit {
...
+
+    // Give the timer its starting state.
+    // (Because of the subscriptions we set up, these initial values should
+    // propagate to Firestore.)
     this.timerService.initialize({
       running: false,
       tickSpeed: 1000,
       currentTime: this.startTime,
       endTime: this.endTime
     });
```
floogulinc, line 242: I feel like this functionality of "joining"/setting up a round should be in its own service, especially the subscribing to timer and pushing updates to Firebase. This would probably fit in the planned `TeacherRoundService` and some of it in the `TeacherSessionService` perhaps.

src/app/large-display/large-display.component.ts
```diff
@@ -111,34 +153,107 @@ export class LargeDisplayComponent implements OnInit, AfterViewInit {
...
-  running: boolean;
+  currentScreen: ScreenId = ScreenId.WaitingToAuthenticate;
+
+  running$: Observable<boolean> = this.timerService.running$.pipe(
+    // We need the explicit type parameters or else tslint can't tell which
+    // overload we're using, and it mistakenly thinks we're invoking the
+    // function in a deprecated manner.
+    startWith<boolean, null>(null),
+    shareReplay(1),
+  );
```
floogulinc, line 164: Why are we setting this up in the component? Is there something extra needed that `running$` from the timer service doesn't provide?

src/app/large-display/large-display.component.ts
```diff
@@ -111,34 +153,107 @@ export class LargeDisplayComponent implements OnInit, AfterViewInit {
...
 
-  constructor(public timerService: TimerService) { }
+  readonly roundPath$ = new BehaviorSubject<RoundPath | null>(null);
```
floogulinc, line 166: This is also something that should be part of a `TeacherRoundService`.

src/app/large-display/large-display.component.ts
helloworld12321: Oh, I didn't realize `TimerService.running$` had a `shareReplay()` on it! Never mind then :) Thanks!

helloworld12321:
(The regressions in the layout should be all fixed now! :slightly_smiling_face:)

helloworld12321:
Roger that, thanks! I'll need to update this branch to incorporate PR #76—I'll do that Monday morning—and then should I maybe ask for one person to re-review the changes?

helloworld12321:
Okay, I know that Audrey's review is a bit stale, but given that it looks like nothing changed when we merged in the Firebase emulator, and given that we don't want to drag this review process out for longer than we have to, I think it makes sense to merge in this branch now. (Sorry if that's the wrong decision!)

PR #74, opened 2020-07-08
Set up Firebase Emulators

floogulinc:
The major change here is that when you are running the dev server through `ng serve` it will be set to use the Firebase emulators for Firestore rather than the live production database. This was configured in the `environment.ts` and `app.module.ts` files. Authentication still happens against the live Firebase instance though. 

We have setup a NPM script to run the emulator with settings that are most applicable to us currently. It is `npm run firebase:emulator`. This currently runs `firebase emulators:start --only firestore --import firebase-test-data`.

You should run the emulators at the same time as running `ng serve`. It provides a graphical web interface that is similar to the real Firebase dashboard. The command will tell you how to access it when you start the emulators.

We have setup a set of test data consisting of the `demo-session` and `demo-round` we already had in production. This was exported to the `firebase-test-data` directory and will be imported with the command described above. If you want to update this test data, make the changes you want in the local Firestore and run `firebase emulators:export firebase-test-data`.


floogulinc:
Very interesting @helloworld12321. Looks like when you export on Windows it uses `\` in some files. See upstream issue https://github.com/firebase/firebase-tools/issues/2421. 

The solution at the moment may be that we have to export on *nix (or maybe we can get away with manually changing the paths). As far as I know Windows actually handles both `/` and `\` as path separators so using `/` may just work for Windows as well.

helloworld12321:
Ooh, nice find on that issue!

I wonder, would it make sense to populate the database programmatically as part of the test suite? (That also would give us a degree of flexibility—for example, we'd be able to test the teacher-side with & without a previous session that they could resume.)

floogulinc:
Yeah for testing the plan indeed is to setup stuff in the database during the test. We just need this basic demo data for local development at the moment. Once we have session and round creation though we won't need to be seeding it.

floogulinc:
So for the moment if you are using Windows and export new data, please don't commit the changed `firebase-test-data\firestore_export\firestore_export.overall_export_metadata` file.

Issue #75, opened 2020-07-09
Convert bottom bar to component


audreylemeur:
The bottom bar in the large display is currently a div. We would like to make it a component so we can use it in other components.

PR #76, opened 2020-07-09
Change bottom bar to component

audreylemeur:
Made a bottom bar component and switched it out in the large display. 

Addresses issue #75.

Co-authored-by: Shawn Reuter <reute078@morris.umn.edu>


src/app/large-display/large-display.component.html
```diff
@@ -6,19 +6,19 @@
...
+  <div left>
     <button mat-raised-button color="accent">Quit Game</button>
   </div>
-  <div class="center">
+  <div center>
     <button mat-mini-fab color="accent"
       [disabled]="running === undefined"
       (click)="toggleTimerRunning()"
-    >
+      >
```
helloworld12321, line 17:
```suggestion
    >
```
Just a quick indentation fix!

helloworld12321: Unless that was intentional on your part, in which case never mind, ignore me :slightly_smiling_face:

PR #80, opened 2020-07-15
Prepare round

audreylemeur:
Note that this pull request depends on pull request #68 being merged into master.


src/app/prepare-round-test/prepare-round-test.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  styleUrls: ['./prepare-round-test.component.scss']
+})
+export class PrepareRoundTestComponent implements OnInit {
+
+  userID: string;
+  testSession: string;
+  testRoundData: FirebaseRound;
+
+  constructor(public firebaseService: FirebaseService) {
+    this.testSession = 'kugTpWqJyrXaJZ4ZB6zE';
```
helloworld12321, line 19: Would it be possible to create this session if it doesn't already exist? (At the moment, you need to create it manually with the Firebase console.)

src/app/session-lobby/session-lobby.component.ts
```diff
@@ -9,13 +11,22 @@ import { TeacherSessionService } from '../teacher-session.service';
...
+  roundData: FirebaseRound;
 
-  constructor(public teacherSessionService: TeacherSessionService) {
+  constructor(public teacherSessionService: TeacherSessionService, public teacherRoundService: TeacherRoundService) {
     this.sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
+    this.roundData = {flowerSpeciesIds: ['asclepias_syriaca', 'coreopsis_palmata'],
+                          status: 'fine',
+                          running: false,
+                          currentTime: 0, };
   }
```
helloworld12321, line 22:
```suggestion
  constructor(public teacherSessionService: TeacherSessionService, public teacherRoundService: TeacherRoundService) {  }
```

You can probably simplify this a bit by assigning these two fields where they're declared.

src/app/session-lobby/session-lobby.component.ts
```diff
@@ -9,13 +11,22 @@ import { TeacherSessionService } from '../teacher-session.service';
 export class SessionLobbyComponent implements OnInit {
 
   sessionID: string;
+  roundData: FirebaseRound;
```
helloworld12321, line 14:
```suggestion

  sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
  roundData = {flowerSpeciesIds: ['asclepias_syriaca', 'coreopsis_palmata'],
                        status: 'fine',
                        running: false,
                        currentTime: 0, };
```

src/app/large-display/large-display.component.html
```diff
@@ -1,52 +1,34 @@
-<ng-container [ngSwitch]="currentScreen">
```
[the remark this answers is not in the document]
floogulinc: I'm not sure why we are removing this stuff since we will need it in the future to have different screens for different parts of a round (pre-gameplay and post-gameplay).

src/app/firebase.service.ts
```diff
@@ -66,6 +66,21 @@ export class FirebaseService {
...
+  /**
+   * Adds round to firestore and sets round to the session's current round
+   * @param sessionID Session ID to which round will be added
+   * @param roundData data for the round in FirebaseRound interface format
+   */
+  public async createRoundInSession(sessionID: string, roundData: FirebaseRound){
+    let round: Promise<DocumentReference>;
+    round = this.firestore.collection('sessions/' + sessionID + '/rounds').add(roundData);
+    this.setCurrentRound({sessionId: sessionID, roundId: (await round).id});
+  }
```
floogulinc, line 78: This logic should be moved to the session service. The Firebase service should only contain the most basic wrappers around Firebase functionality, so in this case `this.firestore.collection('sessions/' + sessionID + '/rounds').add(roundData);` would be abstracted into its own function in the Firebase service and `createRoundInSession` should be moved to the teacher session (or maybe round) service and get the session from the current session subject. Additionally, `createRoundInSession` should return a promise of the result of setting the current round so that it can be waited for.

src/app/session-lobby/session-lobby.component.ts
```diff
@@ -9,13 +11,22 @@ import { TeacherSessionService } from '../teacher-session.service';
...
 
-  constructor(public teacherSessionService: TeacherSessionService) {
+  constructor(public teacherSessionService: TeacherSessionService, public teacherRoundService: TeacherRoundService) {
     this.sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
+    this.roundData = {flowerSpeciesIds: ['asclepias_syriaca', 'coreopsis_palmata'],
+                          status: 'fine',
+                          running: false,
+                          currentTime: 0, };
```
floogulinc, line 21: We might want to have a separate interface for how we store the round "prototype". We will want to be able to specify the flower species in the round using direct references to the objects and also don't need to include things like `running: false`. I would suggest a round preset type that looks something like this:
```ts
interface PresetRound {
  flowerSpecies: FlowerSpecies[];
  startTime: TimePeriod;
  endTime: TimePeriod;
  tickSpeed: number;
  bees: {
    species: BeeSpecies;
    chance: number;
  }
}
```

src/app/teacher-round.service.ts
```diff
@@ -53,7 +53,7 @@ export class TeacherRoundService {
     // Eventually, we'll create a new round, but for the moment, we'll just use
     // this one.
     this.roundPath$.next(demoRoundPath);
```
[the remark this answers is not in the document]
floogulinc: I feel like it makes the most sense for the round path in this service to just come from Firebase like it does in the student equivalent.

src/app/firebase.service.ts
helloworld12321: Oh, even better, yeah!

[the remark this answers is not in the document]
helloworld12321: (Update: Paul has a pretty good suggestion in his review!)

src/app/prepare-round-test/prepare-round-test.component.ts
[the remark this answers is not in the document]
audreylemeur: Since this is a test that doesn't really need to exist anymore, I'll just leave it.

src/app/firebase.service.ts
```diff
@@ -66,6 +66,21 @@ export class FirebaseService {
...
+
+  public setCurrentRound(roundPath: RoundPath){
```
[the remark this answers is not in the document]
audreylemeur: The current round may not exists if the session has never used a round before.

src/app/large-display/large-display.component.html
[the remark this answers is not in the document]
audreylemeur: The screen is redundant with the session lobby.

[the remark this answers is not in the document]
helloworld12321: Hmm... what about about something like this?

```html
  <ng-container *ngSwitchCase="ScreenId.Lobby">
    <app-session-lobby></app-session-lobby>
  </ng-container>
```

(or I suppose you could simplify that to)
```html
  <app-session-lobby *ngSwitchCase="ScreenId.Lobby"></app-session-lobby>
```

That would fix the redundancy problem w/o having to route to different pages. (The main benefit is that it's easier to manage application state if it's all on one page. Otherwise, the teacher could return to the session lobby by pressing the back button, which is a pain to have to deal with.)

src/app/firebase.service.ts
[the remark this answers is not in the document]
helloworld12321: Oh, right, good point!

src/app/large-display/large-display.component.html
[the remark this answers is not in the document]
floogulinc: Yes but the session lobby shouldn't be its own router it should be inserted in the large display component and we will have a round lobby and such in the future as well.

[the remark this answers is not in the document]
helloworld12321: @floogulinc Could you say more about what you mean by "shouldn't be its own router"?

src/app/session-lobby/session-lobby.component.ts
audreylemeur: Yes, but probably in a separate pull request.

audreylemeur:
If you want to test out the session lobby in the large display, you will have to delete the current round id in firebase.

floogulinc:
Should we add a working "quit round" button to the bottom bar in the large display when in a round that just deletes the current round id in the session?

helloworld12321:
Yeah, that would be really cool! We briefly looked into that today, but got super lost, so we moved on to more pressing issues. I would appreciate if such a thing existed, though!

src/app/session-lobby/session-lobby.component.ts
```diff
@@ -8,14 +10,21 @@ import { TeacherSessionService } from '../teacher-session.service';
...
-  sessionID: string;
 
-  constructor(public teacherSessionService: TeacherSessionService) {
-    this.sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
-  }
+  sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
+  roundData = {flowerSpeciesIds: ['asclepias_syriaca', 'coreopsis_palmata'],
+                        status: 'fine',
+                        running: false,
+                        currentTime: 0, };
```
helloworld12321, line 18:
```suggestion
                        currentTime: teacherRoundService.startTime.time, };
```

We might want to use the demo start time in the demo round data, instead of just putting in `0`. (Credit to @floogulinc for catching this!)

src/app/session-lobby/session-lobby.component.ts
```diff
@@ -8,14 +10,21 @@ import { TeacherSessionService } from '../teacher-session.service';
...
-  sessionID: string;
 
-  constructor(public teacherSessionService: TeacherSessionService) {
-    this.sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
-  }
+  sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
+  roundData = {flowerSpeciesIds: ['asclepias_syriaca', 'coreopsis_palmata'],
+                        status: 'fine',
+                        running: false,
+                        currentTime: teacherRoundService.startTime.time, };
```
ErikRauer, line 18:
```suggestion
                        currentTime: this.teacherRoundService.startTime.time, };
```

src/app/large-display/large-display.component.html
```diff
@@ -1,32 +1,22 @@
...
-    </div>
-    
-    <app-bottom-bar>
-      <div right>
-        <button mat-raised-button color="accent" (click)="startRound()">
-          Start Round
-        </button>
-      </div>
-    </app-bottom-bar>
+    <app-session-lobby></app-session-lobby>
```
ErikRauer, line 3: Since we're doing this, we should probably remove the routing to the Session Lobby Component (Especially because the Start Round button doesn't seem to work when the Session Lobby Component is not part of the Large Display Component)

src/app/session-lobby/session-lobby.component.ts
helloworld12321: Oops, thanks! :relaxed: :sheep: <sub>(that means "sheepish")</sub>

src/app/firebase.service.ts
```diff
@@ -85,6 +100,7 @@ export class FirebaseService {
   }
 
   /**
+\
```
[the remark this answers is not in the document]
helloworld12321: Oops :P Thanks!

src/app/large-display/large-display.component.html
helloworld12321: Roger that!

src/app/large-display/large-display.component.html
```diff
@@ -45,7 +35,12 @@
       </div>
 
       <div right>
-        <button mat-raised-button color="accent">Skip Round</button>
+        <button
+          mat-raised-button
+          color="accent"
+        >
+          Skip Round
+        </button>
```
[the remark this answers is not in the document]
helloworld12321: Not particularly—I'll change that later today :)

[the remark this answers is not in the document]
helloworld12321: (Done!)

helloworld12321:
Okay, that should take care of those. Thanks, Erik! :slightly_smiling_face:

src/app/session-lobby/session-lobby.component.ts
```diff
@@ -8,14 +10,21 @@ import { TeacherSessionService } from '../teacher-session.service';
...
 export class SessionLobbyComponent implements OnInit {
 
-  sessionID: string;
 
-  constructor(public teacherSessionService: TeacherSessionService) {
-    this.sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
-  }
+  sessionID = 'demo-session'; // Temporary until a way to get the session is implemented
+  roundData = {flowerSpeciesIds: ['asclepias_syriaca', 'coreopsis_palmata'],
+                        status: 'fine',
```
floogulinc, line 16: I know we haven't defined what a round `status` is, but what does "fine" mean here?

src/app/teacher-round.service.ts
```diff
@@ -42,29 +50,23 @@ export class TeacherRoundService {
...
    * active round, set its initial state, and hook up the TimerService so that
    * when the timer ticks, the round updates.
-   *
-   * TODO: As of this iteration, this function just re-uses a single demo
-   * round and re-populates it with data, rather than creating a new round.
-   * We also don't bother with marking it as the current round.
    */
   // In the future, we might get sessionId from a TeacherSessionService, rather
   // than passing it in as a parameter.
   startNewRound(sessionId: string, roundData: FirebaseRound): void {
```
floogulinc, line 56: Can't we get the session ID from the session service rather than passing it in here?

src/app/teacher-round.service.ts
```diff
@@ -42,29 +50,23 @@ export class TeacherRoundService {
...
-    // this one.
-    this.roundPath$.next(demoRoundPath);
-    this.firebaseService.setRoundData(demoRoundPath, roundData);
-  }
+    let round: Promise<RoundPath>;
+    round = this.firebaseService.createRoundInSession(sessionId, roundData);
+    round.then(roundPath => {
+      this.firebaseService.setCurrentRound(roundPath);
+      this.teacherSessionService.currentRoundPath$.next(roundPath);
+    });
```
floogulinc, line 62:
```suggestion
    this.firebaseService.createRoundInSession(sessionId, roundData).then(roundPath => {
      this.firebaseService.setCurrentRound(roundPath);
      this.teacherSessionService.currentRoundPath$.next(roundPath);
    });
```

Also it may be useful to return this promise.

src/app/teacher-round.service.ts
```diff
@@ -42,29 +50,23 @@ export class TeacherRoundService {
...
-    // Eventually, we'll create a new round, but for the moment, we'll just use
-    // this one.
-    this.roundPath$.next(demoRoundPath);
-    this.firebaseService.setRoundData(demoRoundPath, roundData);
-  }
+    let round: Promise<RoundPath>;
+    round = this.firebaseService.createRoundInSession(sessionId, roundData);
+    round.then(roundPath => {
+      this.firebaseService.setCurrentRound(roundPath);
+      this.teacherSessionService.currentRoundPath$.next(roundPath);
```
floogulinc, line 61: You might want to set the `currentRoundPath$` after you know that `firebaseService.setCurrentRound` succeded.

src/app/teacher-session.service.ts
```diff
@@ -31,6 +40,12 @@ export class TeacherSessionService {
     shareReplay(1),
   );
 
+  /**
+   * An observable of the current round's session ID and round ID.
+   * Emits null if the student is not in a session or the round is not set on the session.
+   */
+  currentRoundPath$ = new BehaviorSubject<RoundPath | null>(null);
```
floogulinc, line 47: Maybe this should be in the round service?

src/app/session-lobby/session-lobby.component.ts
helloworld12321: Oh, that was supposed to be a blatant placeholder—something that obviously isn't production data 😛

src/app/teacher-session.service.ts
helloworld12321: That would work just fine, but the current round kind of feels like a property of the session.  🤷🏻‍♂️ (That's also how `StudentRoundService` treats it.)

floogulinc: I do get that but it did feel weird that it was being *set* by something in the round service.

helloworld12321: Oh, I suppose 😄

audreylemeur: I agree with Joe's point about the current round being the property of the session.

src/app/teacher-round.service.ts
```diff
@@ -42,29 +50,27 @@ export class TeacherRoundService {
...
   // In the future, we might get sessionId from a TeacherSessionService, rather
   // than passing it in as a parameter.
-  startNewRound(sessionId: string, roundData: FirebaseRound): void {
-    // Eventually, we'll create a new round, but for the moment, we'll just use
-    // this one.
-    this.roundPath$.next(demoRoundPath);
-    this.firebaseService.setRoundData(demoRoundPath, roundData);
-  }
+  startNewRound(roundData: FirebaseRound): void {
+    let round: Promise<RoundPath>;
```
floogulinc, line 57: Why are we creating a variable for this when we never use it to return the promise?

src/app/firebase.service.ts
```diff
@@ -66,6 +66,23 @@ export class FirebaseService {
...
+   */
+  public createRoundInSession(sessionId: string, roundData: FirebaseRound): Promise<RoundPath> {
+    return this.firestore.collection('sessions/' + sessionId + '/rounds').add(roundData).then(doc =>
+      ({sessionId, roundId: doc.id})
+    );
+  }
+
+  public setCurrentRound(roundPath: RoundPath): Promise<RoundPath>{
+    return this.firestore.collection('sessions').doc(roundPath.sessionId).update({currentRoundId: roundPath.roundId}).then(doc =>
+      (roundPath)
```
floogulinc, line 82: I don't know how useful it is to just return the same `roundPath` that was input to the function since you would already know it.

ErikRauer: What would you prefer we return instead?

audreylemeur: What would you suggest instead?

helloworld12321: Could we do a `Promise<void>`?

Something like

```ts
  public async setCurrentRound(roundPath: RoundPath): Promise<void> {
    await this.firestore.collection('sessions').doc(roundPath.sessionId)
      .update({currentRoundId: roundPath.roundId});
  }
```

(equivalently,)

```ts
  public setCurrentRound(roundPath: RoundPath): Promise<void> {
    return this.firestore.collection('sessions').doc(roundPath.sessionId)
      .update({currentRoundId: roundPath.roundId}).then(() => {});
  }
```

src/app/teacher-round.service.spec.ts
```diff
@@ -15,15 +17,35 @@ describe('TeacherRoundService', () => {
...
+
     const mockFirebaseService = jasmine.createSpyObj<Partial<FirebaseService>>(
       'firebaseService',
-      ['updateRoundData', 'setRoundData'],
+      ['updateRoundData', 'setRoundData', 'createRoundInSession', 'setCurrentRound'],
     );
 
+    mockFirebaseService.createRoundInSession.and.callFake(() => {
+      const fakeRoundPath = {sessionId: fakeSessionId, roundId: 'demo-round'};
+      mockCurrentRoundPath$.next(fakeRoundPath);
```
ErikRauer, line 35: @helloworld12321 Is there a reason you set the mockCurrentRoundPath$ here? The actual createRoundInSession in FirebaseService does not change currentRoundPath$

src/app/firebase.service.ts
floogulinc: You could just return whatever the response is from Firebase, you are always free to ignore it with a `() => {}` kind of thing.
```ts
public setCurrentRound(roundPath: RoundPath) {
    return this.firestore.collection('sessions').doc(roundPath.sessionId)
      .update({currentRoundId: roundPath.roundId});
  }
```

src/app/teacher-round.service.spec.ts
helloworld12321: Ah, good catch, thanks! I'll check that out later today.

src/app/firebase.service.ts
helloworld12321: Oh, yeah, nifty! :relaxed:

src/app/teacher-round.service.spec.ts
helloworld12321: Fixed!

src/app/teacher-round.service.ts
```diff
@@ -42,29 +50,24 @@ export class TeacherRoundService {
...
-    // this one.
-    this.roundPath$.next(demoRoundPath);
-    this.firebaseService.setRoundData(demoRoundPath, roundData);
-  }
+  startNewRound(roundData: FirebaseRound): void {
+    this.teacherSessionService.sessionId$.pipe(take(1)).subscribe(sessionId => {
+      this.firebaseService.createRoundInSession(sessionId, roundData).then(roundPath => {
+        this.firebaseService.setCurrentRound(roundPath).then(() => {
+          this.teacherSessionService.currentRoundPath$.next(roundPath); });
+      });
```
floogulinc, line 61: The indentation and bracketing here seems a bit off

helloworld12321:
```suggestion
    this.teacherSessionService.sessionId$.pipe(take(1)).subscribe(sessionId => {
      this.firebaseService.createRoundInSession(sessionId, roundData).then(roundPath => {
        this.firebaseService.setCurrentRound(roundPath).then(() => {
          this.teacherSessionService.currentRoundPath$.next(roundPath);
        });
      });
```

How about this?

helloworld12321:
I mean, I'm not particularly worried about the formatting :upside_down_face: It's not like the checkstyle is complaining or anything.

helloworld12321:
(And, ultimately, the checkstyle is the final arbiter of these things.)

PR #89, opened 2020-07-23
Assign bees

ErikRauer:
Assign each student a bee when starting the a new round. 

This pull request depends on #80


src/app/teacher-round.service.ts
```diff
@@ -70,4 +88,86 @@ export class TeacherRoundService {
...
+
+    // Randomly assign the leftover students to a bee species
+    for (let i = currentStudent; i < shuffledStudents.length; i++) {
+      const beeIndex = Math.floor(Math.random() * demoBees.length);
+      this.firebaseService.addStudentToRound(shuffledStudents[i].id, path,
+        {beeSpecies: beeList[beeIndex].id});
+    }
+  }
+
+  shuffleArray(array: SessionStudentData[]): SessionStudentData[] {
```
helloworld12321, line 163:
```suggestion
  shuffleArray<T>(array: T[]): T[] {
```

This would be a nifty place to use generics, by the way! :smiley: (That way, you can pass in both `SessionStudentData[]` either `BeeSpecies[]`, depending on what you want.)

src/app/teacher-round.service.ts
```diff
@@ -160,7 +160,7 @@ export class TeacherRoundService {
     }
   }
 
-  shuffleArray(array: SessionStudentData[]): SessionStudentData[] {
+  shuffleArray(array: any[]): any[] {
```
helloworld12321, line 163:
```suggestion
  shuffleArray(array: any[]): any[] {
```

src/app/teacher-round.service.ts
```diff
@@ -70,4 +88,86 @@ export class TeacherRoundService {
...
+
+    // Randomly assign the leftover students to a bee species
+    for (let i = currentStudent; i < shuffledStudents.length; i++) {
+      const beeIndex = Math.floor(Math.random() * demoBees.length);
+      this.firebaseService.addStudentToRound(shuffledStudents[i].id, path,
+        {beeSpecies: beeList[beeIndex].id});
+    }
+  }
+
+  shuffleArray(array: any[]): any[] {
```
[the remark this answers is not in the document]
floogulinc: Yes, this would be a very good idea to do.

src/app/round.ts
```diff
@@ -7,6 +7,7 @@ import { TimePeriod } from './time-period';
  */
 export interface FirebaseRound {
   flowerSpeciesIds: string[];
+  beeSpeciesIds?: string[];
```
floogulinc, line 10: Why are we adding the array of bee species IDs to the firebase round?

helloworld12321: Oh, good catch!

audreylemeur: I think this was a temporary measure to set the bees until the templates are added.

.vscode/settings.json
```diff
@@ -1,5 +0,0 @@
-{
-    "cSpell.words": [
-        "grayscale"
-    ]
-}
```
floogulinc, line 5: Why are we deleting this file?

ErikRauer: Oops, that was an accident.

floogulinc: I went ahead and restored it (I would have asked but I didn't realize the "restore file" button would just do it rather than asking me).

helloworld12321:
Oof, looks like there's a problem in prepare-round-test:

```
12:47:02.274 | ERROR in src/app/prepare-round-test/prepare-round-test.component.ts:22:27 - error TS2322: Type '{ flowerSpeciesIds: string[]; beeSpeciesIds: string[]; status: string; running: false; currentTime: number; }' is not assignable to type 'FirebaseRound'.
12:47:02.274 | Object literal may only specify known properties, and 'beeSpeciesIds' does not exist in type 'FirebaseRound'.
12:47:02.275 | 22                           beeSpeciesIds: [allBeeSpecies.apis_mellifera.id],
12:47:02.275 | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
12:47:02.324 | Error: Command "ng build --prod" exited with 1
```

PR #91, opened 2020-07-23
Create file folders for better organization

reuterShawn:
Created the following folders called components, components-test, services,  and pages. We also could have a folder containing all of the interfaces if necessary.


reuterShawn:
Here is a [link ](https://docs.google.com/document/d/1OOHVjGm2vfJmpZhN98ZCAutcOq8bocC1T1lUbWkHr7s/edit )to the google doc with the file structure. We could also make a folder for all of the files with  an interface if we feel it is necessary.

angular.json
```diff
@@ -148,5 +148,8 @@
       }
     }
   },
-  "defaultProject": "buzz-about"
-}
+  "defaultProject": "buzz-about",
+  "cli": {
+    "analytics": false
+  }
```
helloworld12321, line 154: Sounds good!

angular.json
```diff
@@ -148,5 +148,8 @@
       }
     }
   },
-  "defaultProject": "buzz-about"
-}
+  "defaultProject": "buzz-about",
+  "cli": {
+    "analytics": false
```
floogulinc, line 153: I don't think we should set this for the entire project. It should be an individual choice.

reuterShawn: Sorry Paul I didn't mean to push that up there was 100 files that were being committed so I did not notice that this was one of them.

angular.json
floogulinc, line 154:
```suggestion
  "defaultProject": "buzz-about"
```

PR #94, opened 2020-08-01
Add round templates

floogulinc:
test page: https://buzz-about-git-templateroundinterface.mn-pollinators.vercel.app/round-template-test


src/app/round-template-test/round-template-test.component.scss
```diff
@@ -0,0 +1,6 @@
+app-flower-layout {
+}
+.template-grid-container {
+  display: grid;
+  grid-template-columns: repeat( auto-fit, minmax(300px, 1fr) );
```
helloworld12321, line 5: Spiffy!

src/app/round-template.ts
```diff
@@ -0,0 +1,108 @@
...
+export interface RoundTemplate {
+  name: string;
+  flowerSpecies: FlowerSpecies[];
+  startTime: TimePeriod;
+  endTime: TimePeriod;
+  tickSpeed: number;
+  bees?: {
+    species: BeeSpecies;
+    weight: number;
+  }[];
```
helloworld12321, line 14: Oh, I see y'all decided to make the weights non-optional?

Cool cool :slightly_smiling_face:

src/app/round-template.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+        expect(flowerSpecies.length).toBe(16);
+      });
+
+      if (bees) {
+        describe('if there are bees', () => {
+          it('has at least 1 bee', () => {
+            expect(bees.length).toBeGreaterThanOrEqual(1);
+          });
+          it('bee weights add to 1', () => {
+            expect(bees.reduce((prev, curr) => prev + curr.weight, 0)).toEqual(1);
```
helloworld12321, line 17:
```suggestion
            expect(bees.reduce((prev, curr) => prev + curr.weight, 0)).toBeCloseTo(1);
```

Since we're doing floating-point math, we probably want to use Jasmine's `toBeCloseTo` matcher (otherwise this test could fail because `0.99999999974 !== 1` or something like that :stuck_out_tongue_winking_eye:)

(By default, `toBeCloseTo` is accurate to two decimal places, but you can change the precision if you want!)

floogulinc:
```suggestion
            expect(bees.reduce((prev, curr) => prev + curr.weight, 0)).toBeCloseTo(1, 10);
```
What do you think about expecting at least 10 decimal places?

Issue #104, opened 2020-08-04
Make a basic nice-looking homepage with host and join buttons


floogulinc:
This should include buttons to host and join a session as well as the LCCMR attribution. This should also move all the links to test pages to a new component at `/test`.

This is related to #7.

PR #107, opened 2020-08-04
Make homepage better & separate test pages

UtkarshKr007:
Closes: #104


floogulinc:
@helloworld12321 I ended up going with the stroked button and also made the buttons full width

image

helloworld12321:
Oh, even better, nice!

Issue #100, opened 2020-08-04
Display a "pollen count" for students based on interactions


floogulinc:
- A student can "hold" up to 3 pollen (3 flower interactions) at a time
- A student can then "deposit" the pollen to their nest by visiting it
- Students can see the pollen they are holding at all times but can only see the pollen in their nest when looking at the nest in AR

We should be able to find pollen counts from querying the interactions collection rather than keeping counts locally.

PR #108, opened 2020-08-05
Display pollen count

audreylemeur:
Closes: #100.


floogulinc:
Here's a note I left in Slack in case you didn't see it:
>It would probably be easier to simply get the entire array of all interactions for the current student ID from Firebase and provide that in the round service as a share replay observable. Then for each of the things like current and total pollen counts just do the filtering and math in each of those observable coming from the main interactions one. It isn't really more efficient or saving anything to try and filter out the flower interactions and stuff in the firebase query.

src/app/student-round.service.ts
```diff
@@ -145,6 +145,53 @@ export class StudentRoundService {
     shareReplay(1),
   );
 
+  interactions$: Observable<Interaction[] | null> =
+  combineLatest([this.sessionService.currentRoundPath$, this.authService.currentUser$]).pipe(
+    switchMap(([path, user]) =>
+      path && user
+        ? this.firebaseService.getStudentInteractions(path, user.uid)
+        : of(null)),
```
floogulinc, line 153: I think we should return an empty array here, that would be consistent with how we handle other collections and arrays and reduces the need for null checks.

floogulinc:
I will wait until the branch is updated to master before reviewing.

src/app/components/play-round/play-round.component.ts
```diff
@@ -67,6 +67,15 @@ export class PlayRoundComponent implements OnInit {
...
 
+  beePollen$: Observable<boolean[]> = this.studentRoundService.currentBeePollen$.pipe(
+    map(pollenCount => {
+      const pollenArray: boolean[] = [false, false, false];
+      for (let i = 0; i < pollenCount; i++) {
+        pollenArray[i] = true;
+      }
+      return pollenArray;
+    })
+  );
```
helloworld12321, line 78: Good stuff!

src/app/services/student-round.service.spec.ts
```diff
@@ -808,6 +882,213 @@ describe('StudentRoundService', () => {
...
+      expectObservable(service.currentNestPollen$).toBe(
+        '1-',
+        values.numbers,
+      );
+    });
+  });
+
+  /**
+   *
+   */
```
helloworld12321, line 1091: Holy cow, you really went to town on the testing! Nice job!

src/app/services/student-round.service.ts
```diff
@@ -145,6 +145,45 @@ export class StudentRoundService {
...
+    ),
+    distinctUntilChanged(),
+  );
+
+  currentBeePollen$: Observable<number> =
+  this.interactions$.pipe(
+    map(interactions => {
+        let currentPollen = 0;
+        for (const interaction of interactions) {
+          if (interaction.barcodeValue === 0) {
```
helloworld12321, line 169: (See the comment on `play-round.component.html`)

src/app/services/student-round.service.ts
```diff
@@ -145,6 +145,45 @@ export class StudentRoundService {
...
+    map(interactions =>
+      interactions.filter(interaction => interaction.barcodeValue >= 1 && interaction.barcodeValue <= 16)
+    ),
+    distinctUntilChanged(),
+  );
+
+  currentBeePollen$: Observable<number> =
+  this.interactions$.pipe(
+    map(interactions => {
+        let currentPollen = 0;
```
helloworld12321, line 167: Would you mind adding a little comment here? (To an observer, it's not immediately obvious that this depends on the order the interactions are sorted in :slightly_smiling_face:)

helloworld12321:
@kklamberty 

Oh, right, this pull-request changes Firebase's configuration (som new indexes got added in the database)
but the production database hasn't got the new configuration yet.
(You can still see it locally, though!)

helloworld12321:
Actually, Paul tells me he just changed the configuration for the production version :slightly_smiling_face: Should work just fine now!

src/app/components/play-round/play-round.component.html
```diff
@@ -1,4 +1,20 @@
...
+      <mat-icon color='accent' *ngIf="pollen">lens</mat-icon>
+      <mat-icon color='accent' *ngIf="!pollen">panorama_fish_eye</mat-icon>
+    </div>
+  </mat-card>
+  <div *ngIf="(this.foundRoundMarker$ | async)?.barcodeValue === 0">
+    <mat-card class='nest-card'>
+      <mat-card-content class='nest-content'>
+        <p>Nest:</p>
+        <h1 class='nest-count'>{{(this.studentRoundService.currentNestPollen$ | async) }}</h1>
+      </mat-card-content>
```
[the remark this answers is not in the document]
audreylemeur: Total pollen is a different number than the pollen in the nest though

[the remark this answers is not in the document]
helloworld12321: Oh, right!

src/app/components/play-round/play-round.component.html
```diff
@@ -10,8 +26,10 @@
   mat-raised-button
   class="interact-button"
   color="primary"
-  *ngIf="this.foundRoundMarker$ | async as marker" [disabled]="!marker.active"
+  *ngIf="this.foundRoundMarker$ | async as marker"
+  [disabled]="!marker.active || ((this.studentRoundService.currentBeePollen$ | async) >= 3 && marker.barcodeValue != 0)"
```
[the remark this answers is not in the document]
floogulinc: I would suggest adding an optional boolean to the RoundMarker type called `isNest` and setting that to true in the observable for the nest marker.

src/app/components/play-round/play-round.component.html
```diff
@@ -1,4 +1,20 @@
-<app-ar-view [markers]="arMarkers$ | async" [debug]="false" (markerStates)="this.onMarkerState($event)"></app-ar-view>
+<app-ar-view #arview [markers]="arMarkers$ | async" [debug]="false" (markerStates)="this.onMarkerState($event)"></app-ar-view>
```
floogulinc, line 1: Why was `#arview` added back? It isn't used anywhere.

src/app/services/firebase.service.ts
```diff
@@ -118,9 +118,8 @@ export class FirebaseService {
    * @param sessionID the ID of the session the students are in
    */
   getStudentsInSession(sessionID: string): Observable<SessionStudentData[]> {
-    return this.angularFirestore
-      .collection('sessions')
-      .doc(sessionID).collection<SessionStudentData>('students')
+    return this.angularFirestore.collection('sessions').doc(sessionID)
+      .collection<SessionStudentData>('students')
       .valueChanges({idField: 'id'});
```
floogulinc, line 123: Do we need to be adding the id field at this point? I don't see it being used anywhere.

src/app/services/student-round.service.ts
```diff
@@ -145,6 +145,45 @@ export class StudentRoundService {
...
+        }
+        return currentPollen;
+    }),
+    distinctUntilChanged(),
+  );
+
+  currentNestPollen$: Observable<number> = combineLatest([this.totalPollen$, this.currentBeePollen$]).pipe(
+    map(([total, bee]) =>
+      total.length - bee
+    ),
```
floogulinc, line 182: I feel like the variable names here are not very descriptive.

src/app/services/student-round.service.ts
```diff
@@ -145,6 +145,45 @@ export class StudentRoundService {
...
+  interactions$: Observable<Interaction[]> =
+  combineLatest([this.sessionService.currentRoundPath$, this.authService.currentUser$]).pipe(
+    switchMap(([path, user]) =>
+      path && user
+        ? this.firebaseService.getStudentInteractions(path, user.uid)
+        : of([])),
+    distinctUntilChanged()
+  );
+
+  totalPollen$: Observable<Interaction[]> = this.interactions$.pipe(
```
floogulinc, line 157: Should this just output a number (the length of the array)?

src/app/services/student-round.service.ts
```diff
@@ -145,6 +145,45 @@ export class StudentRoundService {
     shareReplay(1),
   );
 
+  interactions$: Observable<Interaction[]> =
+  combineLatest([this.sessionService.currentRoundPath$, this.authService.currentUser$]).pipe(
+    switchMap(([path, user]) =>
+      path && user
+        ? this.firebaseService.getStudentInteractions(path, user.uid)
+        : of([])),
+    distinctUntilChanged()
```
floogulinc, line 154: Is `distinctUntilChanged()` actually doing anything here for an array without specific checks for equality?

src/app/services/student-round.service.ts
```diff
@@ -145,6 +145,45 @@ export class StudentRoundService {
...
+  );
+
+  totalPollen$: Observable<Interaction[]> = this.interactions$.pipe(
+    map(interactions =>
+      interactions.filter(interaction => interaction.barcodeValue >= 1 && interaction.barcodeValue <= 16)
+    ),
+    distinctUntilChanged(),
```
floogulinc, line 161: Again, is `distinctUntilChanged()` really doing anything here for this array? If this observable output just a number then it would work as is.

src/app/services/student-round.service.ts
helloworld12321: Oh, good catch. Looks like it's just checking reference identity at the moment.

src/app/components/play-round/play-round.component.html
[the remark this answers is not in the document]
ErikRauer: I do agree that it needs to be clearer. Maybe something like "Nest Pollen" or "Pollen in Nest" would be better?

src/app/services/firebase.service.ts
ErikRauer: If I remember correctly we were using it when assigning bees at the start of a round

src/app/components/play-round/play-round.component.html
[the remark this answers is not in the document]
ErikRauer: That's a good idea. I think I'll add a similar boolean to interactions.

src/app/components/play-round/play-round.component.html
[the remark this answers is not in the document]
helloworld12321: Oh, I like "Pollen in Nest"!

ErikRauer:
This pull request is ready to be re-reviewed

src/app/components/play-round/play-round.component.html
```diff
@@ -10,7 +21,8 @@
   mat-raised-button
   class="interact-button"
   color="primary"
-  *ngIf="this.foundRoundMarker$ | async as marker" [disabled]="!marker.active"
+  *ngIf="this.foundRoundMarker$ | async as marker"
+  [disabled]="!marker.active || ((this.studentRoundService.currentBeePollen$ | async) >= 3 && marker.barcodeValue != 0)"
```
helloworld12321, line 25: Just one more, sorry!

```suggestion
  [disabled]="!marker.active || ((this.studentRoundService.currentBeePollen$ | async) >= 3 && !marker.isNest)"
```

ErikRauer: Oops, I didn't see that one. Good catch!

src/app/services/student-round.service.ts
```diff
@@ -145,14 +145,61 @@ export class StudentRoundService {
...
   /**
    * Records an interaction with a specific barcode value.
    *
    * @param barcodeValue the barcode value to submit in the interaction
+   * @param isANest whether the barcode corresponds to a student's nest
    */
-  interact(barcodeValue: number) {
+  interact(barcodeValue: number, isANest?: boolean) {
+    let isNest: boolean;
+    isANest ? isNest = true : isNest = false;
```
floogulinc, line 199: What's going on here? I assume you wanted to allow `isANest` to be optional, in that case you can just set a default value.
```suggestion
  interact(barcodeValue: number, isNest: boolean = false) {
```

But I don't think it even has to be an optional thing since `interact` is only called in one place that is already providing the boolean.

src/app/services/student-round.service.ts
```diff
@@ -145,14 +145,61 @@ export class StudentRoundService {
...
+        ? this.firebaseService.getStudentInteractions(path, user.uid)
+        : of([]))
+  );
+
+  totalPollen$: Observable<number> = this.interactions$.pipe(
+    map(interactions =>
+      interactions.filter(interaction => !interaction.isNest)
+    ),
+    map(interactions =>
+      interactions.length)
```
floogulinc, line 161: The two maps here are redundant.
```suggestion
    map(interactions => interactions.filter(interaction => !interaction.isNest).length)
```

src/app/services/student-round.service.ts
ErikRauer: It was meant to act as sort of a default value, I just didn't realize you could simply set one like that. This implementation is definitely a lot cleaner.

ErikRauer:
Sorry about making you re-submit your review over and over again Joe

helloworld12321 (dismissed):
🤣 No worries!

floogulinc (dismissed):

src/app/components/play-round/play-round.component.html
```diff
@@ -10,7 +21,8 @@
   mat-raised-button
   class="interact-button"
   color="primary"
-  *ngIf="this.foundRoundMarker$ | async as marker" [disabled]="!marker.active"
+  *ngIf="this.foundRoundMarker$ | async as marker"
+  [disabled]="!marker.active || ((this.studentRoundService.currentBeePollen$ | async) >= 3 && !marker.isNest)"
```
helloworld12321, line 25: Seems like everything works out just fine with the security rules!

There is one weird thing, though. At the moment, the "interact" button depends on the `currentPollenCount$` observable to know whether it should be enabled, but that observable takes a brief moment to resolve, and until it gives back a value, the "interact" button is enabled; If you click it during that window of time, you can acquire more than 3 pollens at a time.

What do you think about something like this &hellip;?

```html
<!--
  Make sure the currentBeePollen$ observable is resolved before we try to
  display the button (so we know whether it should be enabled or not).
-->
<ng-container *ngIf="{ count: (this.studentRoundService.currentBeePollen$ | async) } as currentBeePollen">
  <button
    mat-raised-button
    class="interact-button"
    color="primary"
    *ngIf="this.foundRoundMarker$ | async as marker"
    [disabled]="!marker.active || currentBeePollen.count >= 3 && !marker.isNest"
    (click)="clickInteract(marker)"
  >
    Visit {{marker.name}}
  </button>
</ng-container>
```

Here, we don't display the interact button until the `currentPollenCount$` observable resolves. It's a bit hacky, though. If you have a better way to do this—or, alternatively, if you think it's not worth the hassle—that would be great!

helloworld12321: For example, if you have a script that clicks a few hundred times really fast, you can do something like this: :upside_down_face:

The student side, where the student has two much pollen, and the pollen counter is filling up half of the screen

helloworld12321 (changes requested):

src/app/components/play-round/play-round.component.html
helloworld12321: By the way, the way that you can reproduce the "button enabled for just a moment" effect is to go to a blooming flower, get 3 pollens, and then quickly move the blooming flower on- and off-camera. When the blooming flower enters the camera's field of view, the button will be enabled for just a second.

floogulinc: Part of the thing here is that in the future we will also want to not allow students to simply interact with the same flower 3 times in fast succession so we will need some kind of debounce based on the last flower interacted with or on the interact button itself.

helloworld12321: Oh, sure, I hear you! That's a bit of a separate issue, though. Even with debouncing, you could still collect 4, 5… pollen by moving around to different flowers and clicking during the brief window when the button is enabled. (My hundreds-of-pollen script would no longer be viable, but that was more of a proof of concept anyway 🙃)

helloworld12321 (approved):
OK, now everything looks good! =) Thanks so much for getting all that implemented!

src/app/components/play-round/play-round.component.html
```diff
@@ -17,13 +17,15 @@ <h1 class="nest-count">{{(this.studentRoundService.currentNestPollen$ | async) }
...
-  mat-raised-button
-  class="interact-button"
-  color="primary"
-  *ngIf="this.foundRoundMarker$ | async as marker"
-  [disabled]="!marker.active || ((this.studentRoundService.currentBeePollen$ | async) >= 3 && !marker.isNest)"
-  (click)="clickInteract(marker)"
->
-  Visit {{marker.name}}
-</button>
+<ng-container *ngIf="{ count: (this.studentRoundService.currentBeePollen$ | async) } as currentBeePollen">
```
floogulinc, line 20: Why is this being made an object? Could it not just be this:
```suggestion
<ng-container *ngIf="(this.studentRoundService.currentBeePollen$ | async) as currentBeePollen">
```
(and change the reference below to remove `.count`)

floogulinc: oh wait I see it's so when the current pollen is 0 it doesn't evaluate as false. Seems kinda hacky though.

kklamberty (commented):
I don't seem to be able to see the pollen count on my device, but I did see something like a white rectangle show up in the upper right corner of my device when I had visited a flower and then visited my nest. What is that supposed to look like now?

helloworld12321 (commented):
~~Looks good!~~

helloworld12321 (commented):
(Had to make a new review because I edited the phrasing of the previous one)

helloworld12321 (approved):
Take 3:
Sorry, I've been clicking the "comment" button instead of the "approve" button :P
I'm not very good at reading words :sheep:

kklamberty (approved):
I think this looks pretty good. I can see the pollen show up when I visit the flowers. However, I noticed something that we will want to address moving forward. 

The pollen collection button is not visible once you have three "pollen". We should provide a message that tells them to visit their nest (or says that they can't collect more pollen until they visit their nest) in place of the button.

On the other hand, a user who visits their nest still sees that button even if they've just visited their nest. Perhaps we should have a message that tells them to go try to collect more pollen? Maybe some kind of message that says "you successfully deposited 3 pollen" or however many they deposited in their nest. The number increased without fanfare and the button still says I can visit my nest. So, that might be problematic... or, at least not as helpful as it should be eventually.

The basics of the mechanics look good, and I think it makes sense to merge this and open a new issue that addresses some potential changes to help clarify what people should do and guide their interaction a bit more.

kklamberty:
Will we need to get the cypress testing check passing before merging it?

Issue #53, opened 2020-06-22
In the large display, make it so that you can play/pause with the space bar


helloworld12321:
Using the space bar for play/pause controls is pretty standard across websites. For the sake of being idiomatic, and allowing teachers to play/pause quickly, we probably want this functionality on the large display page.

helloworld12321:
Given that it's pretty tricky to do this accessibly, I'm going to mark it as `wontfix`, at least for now. (If someone wants to take another crack at it later, feel free to reopen this issue!)

PR #109, opened 2020-08-06
Pause/Play the timer in large display with spacebar

UtkarshKr007:
Closes: #53 

Waiting for https://github.com/mn-pollinators/buzz-about/pull/91 to be merged first.

### DIFF
--- a/src/app/pages/large-display/large-display.component.ts
+++ b/src/app/pages/large-display/large-display.component.ts
@@ -10,6 +10,7 @@ import { take } from 'rxjs/operators';
 import { TeacherRoundService } from '../../services/teacher-round.service';
 import { TeacherSessionService } from '../../services/teacher-session.service';
 import { ActivatedRoute } from '@angular/router';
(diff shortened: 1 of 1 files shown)


src/app/pages/large-display/large-display.component.ts
```diff
@@ -90,6 +92,7 @@ export class LargeDisplayComponent implements OnInit {
     this.teacherRoundService.endRound();
   }
 
+  @HostListener('window:keyup.Space')
```
floogulinc, line 95: We're going to want to be careful with this. This will react to *all* spacebar events in the large display component and this probably includes when typing in a field. It is not unreasonable that we will have things like text fields in the large display and this will play/pause the round anytime the user hits space while typing. This is also likely an accessibility issue because users may want to navigate the app with their keyboard and use space to select things.

Also this is another reason we should split the "during round" stuff into its own component.

UtkarshKr007:
This was an enhancement feature, we are going to move ahead with more prominent features for now.

PR #115, opened 2020-08-07
Add a bunch of words to the cSpell dictionary and fix some typos

floogulinc:


floogulinc:
> Nice idea. With a shared dictionary, we don't have to add the words to our personal dictionary every time. The words added are case-insensitive right or is there a specific reason for having Firesotre & LCCMR capitalized?

They seem to be case-insensitive. The reason a couple of them were capitalized is just because that's the way the string was that I added it from in the code.

Issue #103, opened 2020-08-04
Allow students to enter a nest barcode when joining a session


floogulinc:
This involves adding a numeric input to the join session page so students can enter their nest barcode ID when joining a session.

PR #121, opened 2020-08-08
Students can enter a nest barcode when joining a session

UtkarshKr007:
This will allow students to enter their barcode value for their nests when they join the session. The input brings up the numeric keypad for both Android & iOS. Autocomplete is also disabled for fields as described in #122 .

Closes: #103 
Closes: #122


src/app/pages/join-session/join-session.component.html
```diff
@@ -10,7 +10,14 @@
         </mat-form-field>
         <mat-form-field appearance="outline" class="form-field">
           <mat-label>Name</mat-label>
-          <input matInput class="name"  formControlName="nameControl">
+          <input matInput class="name" formControlName="nameControl">
+        </mat-form-field>
+        <mat-form-field appearance="outline" class="form-field">
+          <mat-label>Nest Barcode</mat-label>
```
floogulinc, line 16: "Barcode" is likely not a good description for this public facing interface. I would suggest something like "Nest marker ID" or "Nest marker number." I think we will be referring to the barcodes as "markers"

src/app/pages/join-session/join-session.component.html
```diff
@@ -10,7 +10,14 @@
...
+          <input matInput inputmode="numeric" pattern="[0-9]*" class="nest" formControlName="nestControl">
+          <mat-error *ngIf="sessionFormGroup.controls['nestControl'].invalid">
+            This number is from 20 to 100 and can be found on the printed nest marker in the upper left corner
```
floogulinc, line 19: I don't think we should include the range here, that is subject to change and doesn't actually matter to the students. Maybe something like "Invalid nest ID: please enter the ID number on your nest marker".

src/app/pages/join-session/join-session.component.html
```diff
@@ -10,7 +10,14 @@
         </mat-form-field>
         <mat-form-field appearance="outline" class="form-field">
           <mat-label>Name</mat-label>
-          <input matInput class="name"  formControlName="nameControl">
+          <input matInput class="name" formControlName="nameControl">
+        </mat-form-field>
+        <mat-form-field appearance="outline" class="form-field">
+          <mat-label>Nest Barcode</mat-label>
+          <input matInput inputmode="numeric" pattern="[0-9]*" class="nest" formControlName="nestControl">
+          <mat-error *ngIf="sessionFormGroup.controls['nestControl'].invalid">
```
floogulinc, line 18: I think it would be a good idea to add a hint here, something like this:
```suggestion
          <input matInput inputmode="numeric" pattern="[0-9]*" class="nest" formControlName="nestControl">
          <mat-hint>The number at the top of your nest marker</mat-hint>
          <mat-error *ngIf="sessionFormGroup.controls['nestControl'].invalid">
```

src/app/pages/join-session/join-session.component.html
floogulinc: Also it should be noted that currently the range of possible barcodes is actually 0-63 but I think we will be switching to a different type of barcode. See #124.

src/app/pages/join-session/join-session.component.ts
```diff
@@ -14,7 +14,9 @@ export class JoinSessionComponent implements OnInit {
 
   sessionFormGroup = new FormGroup({
     nameControl: new FormControl('', Validators.required),
-    sessionControl: new FormControl('', Validators.required)
+    sessionControl: new FormControl('', Validators.required),
+    nestControl: new FormControl('', [Validators.required, Validators.min(18), Validators.max(127),
```
floogulinc, line 18: This should still be 20-120, it seems nice to have a max of exactly 100 nests/students.
```suggestion
    nestControl: new FormControl('', [Validators.required, Validators.min(20), Validators.max(120),
```

src/app/pages/join-session/join-session.component.html
```diff
@@ -10,7 +10,15 @@
         </mat-form-field>
         <mat-form-field appearance="outline" class="form-field">
           <mat-label>Name</mat-label>
-          <input matInput class="name"  formControlName="nameControl">
+          <input matInput class="name" formControlName="nameControl">
+        </mat-form-field>
+        <mat-form-field appearance="outline" class="form-field">
+          <mat-label>Nest marker ID</mat-label>
+          <input matInput inputmode="numeric" pattern="[0-9]*" class="nest" formControlName="nestControl">
```
floogulinc, line 17: I feel like we could make this pattern more specific if we're going to use it at all. Something like this would match exactly 20-120: `[2-8][0-9]|9[0-9]|1[01][0-9]|120`

floogulinc: Also we should mark all the input elements here with the `required` attribute.

UtkarshKr007: Having the inputs be required, only allowing numbers in the nest ID field and checking the value is between 20 and 120 is done by Angular Form Controls' Validators. 
`sessionFormGroup = new FormGroup({
    nameControl: new FormControl('', Validators.required),
    sessionControl: new FormControl('', Validators.required),
    nestControl: new FormControl('', [Validators.required, Validators.min(20), Validators.max(120),
      Validators.pattern('^[0-9]*$')])
  });`

 The `pattern="[0-9]*"` is used to bring up the numerical keypad on iOS devices.

floogulinc: 1. The pattern is not needed for the iOS keyboard, it does that just from the `inputmode`
2. It's still nice to hint to the browser that the field is required.

UtkarshKr007: `inputmode `only bring up the numeric keyboard for iOS 12.2+ which was only available after March 2019. Experimenting on browserstack, there are a couple iOS devices that don't bring up the right keyboard just using `inputmode` and need the `pattern` attribute. We might as well support the most devices possible considering elementary/middle school kids often have hand-me downs or older models anyways.

I can see adding a required attribute for the browsers, but I rather keep the pattern the way it is in both Validator & inputmode just because it's more flexible. If we change the range in the future, we can just change the min and max Validator instead of change those and the regex.

floogulinc: I would prefer to add the `required` attribute to the inputs if we can.

PR #123, opened 2020-08-08
Record teacher event

UtkarshKr007:
Records teacher events during a round to Firebase.  Closes: #112 

Initial Comment:
Currently, only records PauseData which has a field for when the round was paused, but it would be interesting to see if we can record the duration of the pause to estimate both frequency and length of interruptions/discussions during a round. 
If there are other teacher round events we want to record, then we can create a interface called TeacherRoundData and have a field just for pauses.

Update: 
Records all Pause & Play occurrences during a round where the boolean 'running' observable is subscribed to in teacher round service. Currently, we record when the event occurred according to a time relative to the game as well a server timestamp.


floogulinc:
A few initial comments. We should have a general `teacherEvents` or `hostEvents` collection for all the events we want to collect, not one specifically for pauses. Then we can have a common event type that has an event type enum (string), time period, and server timestamp. It could also optionally contain any other arbitrary key value pairs for extra data for particular events.

src/app/pages/large-display/large-display.component.ts
```diff
@@ -70,7 +67,7 @@ export class LargeDisplayComponent implements OnInit {
   toggleTimerRunning() {
     this.timerService.running$.pipe(take(1)).subscribe(running => {
       if (running === true) {
-        this.teacherRoundService.pause();
+        this.teacherRoundService.addHostRoundEvent('Pause');
       }
```
floogulinc: I'd suggest we have two events, `pause` and `play` or something and do something like this:
```ts
this.teacherRoundService.addHostRoundEvent(running ? EventType.Play : EventType.Pause);
```

But also this shouldn't be done here on the button because that won't capture all pause/play events. I think it should probably happen in the `TeacherRoundService` where the paused observable is subscribed to, though there it will also add a single initial event at the start of the round, but that's probably ok.

floogulinc: Alternatively we could have just a single event type for changes in `running` state and have additional data on the event for if running became true or false. But that would probably be more complicated.

src/app/services/teacher-round.service.ts
```diff
@@ -25,9 +23,15 @@ export class TeacherRoundService {
...
+    // Also, record pause & play events based on state of running observable from timer service and
+    // the time of its occurrence.
+    combineLatest([this.teacherSessionService.currentRoundPath$, this.timerService.running$, this.timerService.currentTime$]).pipe(
       filter(([roundPath]) => roundPath !== null),
-    ).subscribe(([roundPath, running]) => {
+    ).subscribe(([roundPath, running, time]) => {
+      this.firebaseService.addHostEvent(
+        roundPath,
+        { eventType: ( running ? HostEventType.Play : HostEventType.Pause ), timePeriod: time.time}
+      );
```
helloworld12321, line 34: Uffda, this is recording the round state every tick :grimacing: I'm sorry about that, Utkarsh!

(Maybe if we put `addHostEvent()` in its own subscription, we could it through a `distinctUntilChanged()` first? :man_shrugging:)

Sorry to be the bearer of bad news!

helloworld12321: (If you want to be extra-especially-cool, you could even capture that behavior in a unit test 😉)

floogulinc: The issue here is subscribing to current time in the combine latest. You want to actually only check the current time when the running state has changed.

helloworld12321: Ah, with a `pipe(take(1))` or something!

firestore-tests/src/firestore.spec.ts
```diff
@@ -212,6 +216,66 @@ describe('Rounds', () => {
...
+    const hostPlayEvent = {
+      eventType: HostEventType.Play,
+      timePeriod: 12
+    };
+
+    beforeEach(async () => {
+      await addStudentToSession(admin, 'Linus', session.id, {
+        name: 'Linus Torvalds',
+        nestBarcode: 91
+      });
```
floogulinc, line 236: This user isn't used in any tests, what is the point of adding it?

floogulinc (changes requested):

firestore-tests/src/firestore.spec.ts
UtkarshKr007: ah, I planned on using the the student, but for got to. I can get rid of it.  Carol and bob are enough show that students can't read or add hostEvents

helloworld12321 (approved):
This looks fantastic—thanks, Utkarsh! :)

floogulinc (approved):

Issue #124, opened 2020-08-09
Choose a different barcode type (and implement it)


floogulinc:
Currently we are using 3x3 AR barcodes. These have some limitations so we should probably change to a different type.

Here is a chart of barcode types (from [artoolkitx](https://github.com/artoolkitx/artoolkitx/wiki/Creating-and-using-square-barcode-markers)):

| Matrix code type                     | Maximum number of markers  | Number of bit errors that can be detected  | Number of bit errors that can be corrected |
| ------------------------------------ | -------------------------- | --------------------- | --------------------- |
| AR\_MATRIX\_CODE\_5x5\_BCH\_22\_7\_7 | 127                        | 6                     | 3                     |
| AR\_MATRIX\_CODE\_4x4\_BCH\_13\_5\_5 | 32                         | 4                     | 2                     |
| AR\_MATRIX\_CODE\_5x5\_22\_12\_5     | 4095                       | 4                     | 2                     |
| AR\_MATRIX\_CODE\_3x3\_HAMMING63     | 8                          | 2                     | 1                     |
| AR\_MATRIX\_CODE\_4x4\_BCH\_13\_9\_3 | 512                        | 2                     | 1                     |
| AR\_MATRIX\_CODE\_3x3\_PARITY65      | 32                         | 1                     | 0                     |
| AR\_MATRIX\_CODE\_3x3                | 64                         | 0                     | 0                     |
| AR\_MATRIX\_CODE\_4x4                | 8192                       | 0                     | 0                     |
| AR\_MATRIX\_CODE\_5x5                | 4194304                    | 0                     | 0                     |

You can see from this chart that the type we are currently using, `AR_MATRIX_CODE_3x3`, only allows up to 64 markers. Since we are using 1-16 for flowers and reserving 0, that leaves only 47 markers left for nests, which for the majority of cases would be enough but could be an issue in particularly large environments. Aside from that the 3x3 barcodes without any parity we are using have no error correction and thus a hamming distance of 0. This means there is a higher chance of a marker being misrecognized as a different ID.

I propose we use the `AR_MATRIX_CODE_5x5_BCH_22_7_7` type as it provides 127 markers, which would give us up to 111 nests and it has the greatest hamming distance between markers and ability to correct for errors in reading. This seems to be the most recommended one in general as well.

floogulinc:
[...]

PR #125, opened 2020-08-09
Marker generator

floogulinc:
https://buzz-about-git-marker-generator.mn-pollinators.vercel.app/markers

This is currently working off the assumption that we change to `AR_MATRIX_CODE_5x5_BCH_22_7_7` type markers, see #124.


floogulinc:
This will need to be changed to use `AR_MATRIX_CODE_4x4_BCH_13_9_3`. See #128

floogulinc:
image

kklamberty (commented):
These all look really nice, Paul! Thanks for making it so flexible. The only kind of marker that doesn't work for me is the 4x6 in landscape. I tried all of the others and they look lovely. I was not able to choose 4x6 and print 4 to a page by downloading the pdf when I made that selection, but that was a limitation of my Preview program. If someone wanted to do that, they would be able to simply choose letter size paper and print 4 to a page in that menu. Since I was a little confused, though, I wonder if teachers might be confused also. There is the 'Save as PDF' option in the print menu. Does it make sense to offer the download menu too? I initially thought I would want to download the pdf since I wasn't going to print them all anyhow, but then I just wasn't sure the best way to proceed. 

I'm also not sure about the option to select "Flower Markers" will be clear. I think we may want to have the option to generate both or either and label it as such. "Flower Markers" and "Nest Markers", where each one is a checkbox/selection thing. Then, if neither option is selected, the print or download buttons would be disabled. I think it's not that intuitive to think about how, as a teacher, I might want to print my flower markers on larger paper to place them around the room and then print my nest markers 4 to a page and give each student one for their own desk location. Currently, that involves two steps:

- To print the flowers larger, I would select "Flower Markers" and then say I have 0 students. 
- To print the nests, I would uncheck "Flower Markers" and then enter my actual number of students. It's not completely intuitive to me that way.

I can imagine that in a world where we have 8 flowers and 1 nest for a student to use without moving around in the shared space, it will be really nice to have the 9 to a page option. In that case, would we want to print the same nest marker for all students? I'm not sure what that looks like, but I've also never looked at another nest marker besides my own as any kind of bee. Do I see "their" nest? Or, what do I see in that case? If we all have the same nest marker, will that be problematic if we are not moving around?

Overall, I think this looks really nice and I appreciate the ability to print things a lot of different ways. I'm not outright "approving" this PR at the moment since the 4x6 landscape option is not working. On the other hand, I would be willing to approve it and put in issues for some of the minor things that would improve this work (in my opinion). Do we just want to say that if you choose 4x6 you only can do portrait?

audreylemeur (changes requested):
The interface for generating the markers looks great. It does seem like there is a lot of white space on the markers.

kklamberty:
> It does seem like there is a lot of white space on the markers.

Paul - was there some kind of logic you were using to determine how much white space to include? I know at some points we have discussed that it would not be ideal to be able to just sit there and look around the room without moving around (in the case where moving around is allowed and encouraged). Audrey - If the markers are very large, I can see two issues - one is that the amount of ink/toner needed to print the markers would be much greater, and the other is that you might not need to move around much to "visit" the flowers. I guess I'm not that concerned about the whitespace, but I don't know if it makes that much sense to support such a large variety of larger paper sizes that I think are pretty uncommon anyhow. If you chose to print a PDF and you had an unusual size of paper, would you really have any trouble without having the "correct" paper size for your printer?

floogulinc:
One thing to keep in mind is that just adding the pdfmake library has increased our main bundle size by about 1mb. It might be a good time to look at code-splitting so we can start only loading these large libraries (pdfmake, arjs, etc) when needed.

floogulinc:
We should look at browser support and potentially disable the print button where it's not supported.

See https://github.com/bpampuch/pdfmake/issues/800

audreylemeur (dismissed):

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,222 @@
...
+      backgroundSVG: this.getBarcode(val).asSVGWithSize(markerSize)
+    }));
+  }
+
+  nestPages(nestCount: number, markerSize: number) {
+    if (nestCount === 0) {
+      return [];
+    }
+    return rangeArray(MIN_NEST_MARKER, MIN_NEST_MARKER + nestCount - 1).map(val =>
+    ({
```
helloworld12321, line 95:
```suggestion
    return rangeArray(MIN_NEST_MARKER, MIN_NEST_MARKER + nestCount - 1).map(val => ({
```
Ditto. Incidentally, I like your use of `rangeArray()` here!

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,222 @@
...
+      value: `${val}`,
+      type: 'Nest',
+      backgroundSVG: this.getBarcode(val).asSVGWithSize(markerSize)
+    }));
+  }
+
+
+  pdfFromPages(
+    pages: Page[],
+    svgHeight: number,
```
helloworld12321, line 105: Would you be so kind as to add a comment saying what units size has? (Am I remembering right that size is measured in points?)

Thank you! ☺️

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,222 @@
...
+      },
+      {
+        width: '*',
+        stack: [
+          'Buzz About',
+          `Markers v${MARKERS_VERSION}`
+        ],
+        alignment: 'right'
+      }
+    ]));
```
helloworld12321, line 144: I really like how the text on the pdf looks—nice job!

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    const columnPages = chunk(markerColumns, numPerPage);
+
+    const backgroundSVGPages = chunk(pages.map(p => p.backgroundSVG), numPerPage);
+
+    if (numPerPage > 1 && pages.length % 2) {
+      columnPages.push([...columnPages.pop(), []]);
+      backgroundSVGPages.push([...backgroundSVGPages.pop(), blankSVG]);
```
helloworld12321, line 152: Okay, that's pretty cool

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,222 @@
...
+  ngOnInit(): void {
+
+  }
+
+
+  getBarcode(value: number) {
+    return new CustomBarcodeMarkerGenerator(BARCODE_TYPE, value);
+  }
+
+  flowerPages(markerSize: number) {
```
helloworld12321, line 81:
```suggestion
  flowerPages(markerSize: number): Page[] {
```
Necessary so the type checker can infer that `type` is a `'Flower' | 'Nest'`

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,222 @@
...
+    return rangeArray(MIN_FLOWER_MARKER, MAX_FLOWER_MARKER).map(val =>
+    ({
+      value: `${val}`,
+      type: 'Flower',
+      backgroundSVG: this.getBarcode(val).asSVGWithSize(markerSize)
+    }));
+  }
+
+  nestPages(nestCount: number, markerSize: number) {
```
helloworld12321, line 90:
```suggestion
  nestPages(nestCount: number, markerSize: number): Page[] {
```

Ditto.

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,222 @@
...
+      markerSize,
+      pageSize,
+      orientationAndNumPerPage
+    );
+  }
+
+  openPDF() {
+    this.makePdf().open();
+  }
+
```
helloworld12321, line 213:
```suggestion
```
Is it okay if we get rid of this method, since it isn't used either? I know you included it for completeness, but I'm just a bit wary about having too much dead code in the project. If you really want to keep it, though, that's fine too! :)

src/app/custom-barcode-marker-generator.ts
```diff
@@ -0,0 +1,36 @@
+
```
helloworld12321, line 1:
```suggestion
```

helloworld12321 (changes requested):
Works like a charm—thanks, Paul!

I've made a few review suggestions—it's mostly
 - rewording one thing in the UI
 - removing some unused methods
 - a bit of documentation
 - narrowing the type-checking on one field
 - some indentation + whitespace
None of that is super critical, though 🙂

(Also, if you disagree with any of the suggestions, I'd be happy to leave things as they are!)

helloworld12321:
Browser compatibility looks pretty good, by the way—I tested on Firefox, desktop Safari, and mobile Safari.

helloworld12321:
In the future, if we need to add anything to the PDF generation, I could see moving it to its own service. Right now, though, I think it's probably fine where it is 🙂

src/app/custom-barcode-marker-generator.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  asSVGDataURIWithSize(size: number) {
+    return (
+      `data:image/svg+xml,${
+      encodeURIComponent(this.asSVGWithSize(size))
+        .replace(/'/g, '%27')
+        .replace(/"/g, '%22')}`
+    );
+  }
```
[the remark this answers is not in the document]
floogulinc: Yeah, this was originally being used to display the barcode in the page but we aren't doing that anymore.

src/app/pages/marker-generator/marker-generator.component.ts
floogulinc, line 105:
```suggestion
    svgHeight: number, // height in points
```

src/app/pages/marker-generator/marker-generator.component.ts
floogulinc: I'd rather keep it there as it signals that is a possible action to anyone looking in the future.

helloworld12321 (commented):
Awesome, thank you! I'm going to give this a quick look over to make sure everything still works, and then I'll approve it :)

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,228 @@
...
+const BARCODE_TYPE = MATRIX_4X4_BCH_1393;
+
+interface Page {
+  /**
+   * The marker number.
+   */
+  value: string;
+  
+  type: 'Flower' | 'Nest';
+
```
helloworld12321, line 36:
```suggestion

  type: 'Flower' | 'Nest';

```
Oof, that trailing whitespace is entirely my fault, sorry. Here's the suggestion to fix it.

src/app/pages/marker-generator/marker-generator.component.ts
```diff
@@ -0,0 +1,228 @@
...
+      }
+    ]));
+
+    const columnPages = chunk(markerColumns, numPerPage);
+
+    const backgroundSVGPages = chunk(pages.map(p => p.backgroundSVG), numPerPage);
+
+    if (numPerPage > 1 && pages.length % 2) {
+      columnPages.push([...columnPages.pop(), []]);
+      backgroundSVGPages.push([...backgroundSVGPages.pop(), blankSVG]);
```
floogulinc, line 158: This logic is actually flawed for `numPerPage > 2` when `pages.length % numPerPage < numPerPage - 1). It should really add an additional `numPerPage - (pages.length % numPerPage)` blank sections. For now this works since the only `numPerpage` possible are 1 and 2.

helloworld12321: Yeah, I saw that too! :) I also figured it was fine.

floogulinc: I have just corrected this in 6e0739bc6281e079807dcb0f575ed8d2dd9caab5

src/app/pages/marker-generator/marker-generator.component.ts
floogulinc: Looks like VSCode automatically corrected it when I was working on 6e0739bc6281e079807dcb0f575ed8d2dd9caab5

helloworld12321 (approved):
Looks fantastic—thanks, Paul.

audreylemeur (approved):

PR #137, opened 2020-09-24
About Page

helloworld12321:


helloworld12321:
In good news, it looks like GitHub doesn't mind if we use the little octocat silhouette next to a link, as long as we don't pretend that we're GitHub-affiliated or anything.

From <https://github.com/logos>:

> ### Do these awesome things
>
> &hellip;
> ✔️ Use the Mark in social buttons to link to your GitHub profile or project 
> &hellip;

audreylemeur (approved):

UtkarshKr007 (approved):
The about page looks really clean. Like the icons. All the links are clickable and lead to the right place.

PR #139, opened 2020-09-29
New development guide and add Cypress to setup guide

floogulinc:
https://github.com/mn-pollinators/buzz-about/blob/dev-docs/DEVELOPMENT.md

https://github.com/mn-pollinators/buzz-about/blob/dev-docs/SETUP.md


DEVELOPMENT.md
```diff
@@ -1,109 +1,87 @@
...
-2. On github.com/settings/tokens, the personal access token has a space after it&mdash;be careful that you don't copy that! :upside_down_face:
-
-![github.com/settings/tokens with the extra space character highlighted][i:The gosh-darn whitespace]
-
-If you run into trouble, you can always fall back to typing the personal access code by hand.
-
-## Installing the Angular command-line interface
-
-Angular provides a CLI with commands for some common operations, like `ng build`, `ng serve`, `ng lint`, and `ng test`. (There's also `ng generate`, which lets you quickly make a new component or service without having to write out the boilerplate every time.)
+You should be able to access the Firebase Emulator Suite page at http://localhost:4000/
```
helloworld12321, line 27: How do you feel about using the word `console` here? I feel like it's a bit more descriptive, in case people didn't already know what the emulator suite page is. :slightly_smiling_face: 

```suggestion
You should be able to access the Firebase Emulator Suite console at http://localhost:4000/
```

DEVELOPMENT.md
```diff
@@ -1,109 +1,87 @@
...
 
-```sh
-npm install -g firebase-tools
+To build the app and run the Cypress end-to-end tests, start the Firebase emulators and run:
+```
+npm run e2e
 ```
 
-The first command that you want to run is probably `firebase login`. This will open a web browser prompting you to log in with your Google account.
+To open Cypress (so you can run the tests interactively), start both the Firebase emulators and the development server and run:
```
helloworld12321, line 55: Is it all right if we switch the order of these phrases to make the development server more prominent? (Otherwise, people might gloss over it and assume that `npm run cy:open` automatically starts the development server.)

(Citation: I did that :upside_down_face:)

```suggestion
To open Cypress (so you can run the tests interactively), start both the development server and the Firebase emulators and run:
```

DEVELOPMENT.md
```diff
@@ -1,109 +1,87 @@
...
+With the Firebase emulators already running, run:
+```
+npm run firestore-specs
+```
 
-### Running the emulator
+To start the Firebase emulators and run through the Firebase tests:
+```
+npm run test-firestore
+```
```
helloworld12321, line 71: Hm... I'm wondering, do you think it would work to only mention `npm run test-firestore` here? (The `firestore-specs` command is sort of redundant with it.)

If not, is there a way to phrase this that more emphasizes the difference between `npm run firestore-specs` and `npm run test-firestore`?

SETUP.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+Now, if you start up a local copy of Buzz About, it should be able to talk to the emulator. Also, if you go to localhost:4000 in a browser, you'll find a nifty little database console!
+
+![The local Firebase console at localhost:4000][i:Local Firebase console]
+
+
+### Setting up Firebase for Cypress
+
+We use [Cypress](https://www.cypress.io/) for our end-to-end testing and use the [cypress-firebase](https://github.com/prescottprue/cypress-firebase) plugin to be able to communicate with Firebase for authentication and the Firebase emulator for Firebase. In addition to having installed `firebase-tools` and used `firebase login`, you will need a Firebase service account for end-to-end testing in Cypress. You can skip this part of the guide if you will not be using Cypress locally.
```
helloworld12321, line 103: What do you think about something like this &hellip;?

```suggestion
We use [Cypress](https://www.cypress.io/) for our end-to-end testing, with the [cypress-firebase](https://github.com/prescottprue/cypress-firebase) plugin to help Cypress talk to Firebase. In addition to the Firebase tools installed above, you will need a Firebase service account for end-to-end testing in Cypress. You can skip this part of the guide if you will not be using Cypress locally.
```

helloworld12321 (changes requested):
Ooh, this is very nice, especially the `DEVELOPMENT.md` that you've written. I've made one or two suggestions where I thought the wording could be more clear, but overall, this looks very good. Thank you! :slightly_smiling_face:

DEVELOPMENT.md
floogulinc: The `firestore-specs` command is useful for more quickly developing re-running specs by keeping the emulator running separately instead of waiting to start it everytime.

helloworld12321: Oh, right! Fair fair.

helloworld12321 (approved):
Looks very nice—thanks, Paul!

kklamberty (approved):

PR #140, opened 2020-10-06
Minor GitHub Actions updates and fixes

floogulinc:
- Set it so we don't install the Cypress binary for runs where it isn't needed.
- Remove the old `ngx-uptodate` workflow, it hasn't worked for a long time now.
- Update `actions/checkout` and `actions/cache` to `v2`


kklamberty (approved):

helloworld12321 (approved):
Looks great, thanks for doing that!

PR #143, opened 2020-10-09
Test creating and joining a session end-to-end

kklamberty:

PR #150, opened 2020-10-20
Fix a bug where all Cypress/Mocha/Chai types were getting imported into app code.

helloworld12321:
Previously, Cypress's types were declared in a .d.ts file in
node_modules, which TypeScript is set up to read from automatically. For
most packages, this wouldn't a problem, since the types would have been
namespaced--you wouldn't actually see them unless you import that
package.

However, Cypress includes Chai and Mocha, which dump their types
directly into the global namespace. This is fine for the E2E tests, but
when we're just writing regular app code, we don't want TypeScript to
think that `expect` and `Assertion` and `it` are defined. (Because they
aren't! 🙃)

This is a particularly bad problem in the `.spec.ts` files, where Chai
and Jasmine are duking it out for control of the global namespace.

Cats and dogs, living together!

We can fix this problem by modifying the `tsconfig.app.json` and
`tsconfig.spec.json` files. These files govern the app code and Karma
tests, respectively, and we can tell them to ignore types coming from
Cypress.

This is still a bit tricky, though, because TypeScript doesn't actually
provide a way to exclude type declarations--there's an 'rule-in' list, but
no 'rule-out' list.

So, instead, for the app code and Karma tests, we provide an *alternate*
`.d.ts` file for Cypress to use--one that's entirely blank. That way, in
the app code an Karma tests, Cypress won't export any types.

See the links below for more information! 🙂

- <https://github.com/microsoft/TypeScript/issues/17042#issuecomment-317589758>
- <https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping>

### DIFF
--- a/no-types.d.ts
+++ b/no-types.d.ts
@@ -1,0 +1,9 @@
+/**
+ * This is a type declaration file that doesn't declare anything. It's useful
+ * for 'masking' certain types that you don't want to include in your project.
+ *
+ * For example, if there's a certain library that you don't want to be typed,
+ * you can tell TypeScript to use this definition file for it instead.
+ *
+ * See: https://github.com/microsoft/TypeScript/issues/17042#issuecomment-317589758
+ */
--- a/tsconfig.app.json
+++ b/tsconfig.app.json
@@ -10,5 +10,10 @@
   ],
   "include": [
     "src/**/*.d.ts"
-  ]
+  ],
+  "paths": {
+    "cypress": [
+      "no-types.d.ts"
+    ]
+  }
 }
--- a/tsconfig.spec.json
+++ b/tsconfig.spec.json
@@ -14,5 +14,10 @@
   "include": [
     "src/**/*.spec.ts",
     "src/**/*.d.ts"
-  ]
+  ],
+  "paths": {
+    "cypress": [
+      "no-types.d.ts"
+    ]
+  }
 }


kklamberty (approved):
Thanks for catching and fixing this issue!

floogulinc:
I have noticed this issue only occurs on the #143 branch but not in `master`.

floogulinc:
It would appear the actual cause of this issue is this import:
https://github.com/mn-pollinators/buzz-about/blob/8d086514b17a0f5929246caf327fef51582c787d/cypress/support/commands.ts#L27

It appears to just be confusing VS Code as the actual unit tests still run fine. There appear to be two simple solutions to this that I think are better than the proposed fix:

1. Remove the `cypress-wait-until` package. We aren't actually using it now so this would probably be fine.
2. Change `cypress/support/commands.ts` to `commands.js` to avoid it using TypeScript at all. This may also be nice since there is an existing error that isn't actually an error we get with it being TypeScript in VS Code:
image

helloworld12321:
Ah, nice debugging, thank you @floogulinc! Yep, if `cypress-wait-until` is being buggy, I'm happy to remove it. (We can always look for an alternative if necessary.)

I'm closing this PR and removing `cypress-wait-until` from #143.

helloworld12321:
As for that `'cypress-firebase'` import, I did some digging, and it looks like that package has a bad `index.d.ts` file: see <https://github.com/prescottprue/cypress-firebase/issues/173>. The workaround is to directly import the directory `'cypress-firebase/lib'`, which skips the `index.d.ts`. I'm pushing a change now! :slightly_smiling_face:

PR #151, opened 2020-10-26
Make the controls on the play-round page scale with the screen size

helloworld12321:
So that, if the students are on an iPad, we can make use of the increased screen real-estate. :slightly_smiling_face: In particular, we're scaling up the interact button, so that it's easy to click, and the pollen counter, so that it's easy to read.

We've also discussed doing away with the interact button in favor of some sort of gestural control; we may end up prototyping that in the future. For the moment, though, we're going to keep maintaining the interact button, just in case we end up preferring it to the gesture controls.

(Edit: note that this PR changes the interact button so that the name of the flower is no longer displayed. If we go with this change, we should probably figure out where to put that text—see issue #153)


helloworld12321:
Some notes from the meeting about things to change:

* Moving the FAB to the right-hand corner
* Plane-landing icon and bee-and-flower-icon → arrow-up icon
* Simplify the scaling code a bit (media queries?)

helloworld12321:
Okay, all fixed! :)
Nexus 6P screenshot
iPad screenshot

src/app/components/play-round/play-round.component.scss
```diff
@@ -18,10 +18,9 @@
 
 .interact-button {
   position: absolute;
-  left:50%;
-  transform: translate(-50%, 0);
-  bottom: 5%;
-  text-transform: uppercase;
+  right: 10px;
+  bottom: 10px;
```
floogulinc, line 22:
```suggestion
  right: 16px;
  bottom: 16px;
```

https://material.io/components/buttons-floating-action-button#specs

src/app/components/play-round/play-round.component.scss
```diff
@@ -37,6 +36,13 @@
   justify-content: space-around;
   align-content: space-around;
   align-items: center;
+
+  // Scale the font's point size up to 6% of the average of the viewport width
+  // and the viewport height. (Never scale it down, though.)
+  // This makes the pollen indicators look pretty good in both landscape and
+  // portrait mode across a range of devices.
+  font-size: max(calc(6 * (1vw + 1vh) / 2), 100%);
+  padding: 0.6em 0.4em;
```
floogulinc, line 45: Do we still want to be scaling this proportionally to width?

floogulinc (changes requested):

src/app/components/play-round/play-round.component.scss
helloworld12321: Ah, nice, thanks!

src/app/components/play-round/play-round.component.scss
helloworld12321: I think so—the FAB only has to be as large as someone's thumb, so we can get away with a one-size-fits-all approach, but the pollen counter needs to be prominent, so we'll want it to take up more area on larger screens.

I'm definitely open to other ways doing that scaling, though!

floogulinc:
I think this will need more refinement in the future but I'll approve what we have now.

floogulinc (approved):

helloworld12321:
Yes, agreed—there's a lot of work left to do here!

UtkarshKr007 (approved):
I checked this out last week, but forgot to leave a review, sorry. I think this is good to be merged in and work with.

PR #157, opened 2020-11-03
Only allow each flower to be visited once per nest cycle (per student)

helloworld12321:


src/app/components/play-round/play-round.component.html
```diff
@@ -17,13 +17,18 @@ <h1 class="nest-count">{{(this.studentRoundService.currentNestPollen$ | async) }
...
+    recentFlowerInteractions: (studentRoundService.recentFlowerInteractions$ | async)
+  } as context"
+>
   <button
     mat-raised-button
     class="interact-button"
     color="primary"
     *ngIf="this.foundRoundMarker$ | async as marker"
-    [disabled]="!marker.active || currentBeePollen.count >= 3 && !marker.isNest"
+    [disabled]="!canVisit(marker, context)"
```
floogulinc, line 31: We really need to avoid using a function call in data binding like this. This function will be called *every* change detection tick and either tank performance or just hang.

helloworld12321: Ah, good to know, thank you very much! Looks like we can fix that problem by using [a pipe instead of a function call](https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496#72eb), since Angular is smarter about when to change-detect a pipe. I'll try to implement that later today! :)

helloworld12321: (Actually, we talked about this at the meeting, and we're probably going to put this information on a field on the `RoundMarker`s!)

helloworld12321: Since that'll involve a bit of refactoring, I probably get to that some other day :)

floogulinc (approved):

UtkarshKr007 (approved):
The code makes sense. I tried it on my phone and can't 'break' the functionality no matter what I try, so great work!!

helloworld12321:
Why thank you! :blush: Thanks for looking at it!

PR #165, opened 2020-12-10
Add join codes

floogulinc:


helloworld12321:
Sorry about the force-push. I made a few commits under the wrong git account—that's my fault.

floogulinc:
We should have an error snackbar for when generating a joincode fails (after the retries).

helloworld12321:
> We should have an error snackbar for when generating a joincode fails (after the retries).

Oh, totally, I'll go do that—thank you!

helloworld12321:
```ts
    }, err => {
      // blah blah snackbar
      this.joinCodeButtonDisabled$.next(false);
    });
```

Yeah, that's probably not what we want in production.

floogulinc:
> ```ts
>     }, err => {
>       // blah blah snackbar
>       this.joinCodeButtonDisabled$.next(false);
>     });
> ```
> 
> Yeah, that's probably not what we want in production.

Haha I totally forgot we left that. oops!

helloworld12321:
Cool, should be all done now!

src/app/join-code.ts
```diff
@@ -0,0 +1,40 @@
...
+export function joinCodesAreEqual(
+  first: JoinCodeWithId | null,
+  second: JoinCodeWithId | null,
+): boolean {
+  if (first === null && second === null) {
+    return true;
+  } else if (first === null && second !== null) {
+    return false;
+  } else if (first !== null && second === null) {
+    return false;
```
floogulinc, line 27: Could this be
```suggestion
  if (first === null || second === null) {
    return first === second;
```

src/app/pages/join-session/join-session.component.html
```diff
@@ -17,7 +20,9 @@
           <input matInput inputmode="numeric" pattern="[0-9]*" class="nest" formControlName="nestControl" required>
           <mat-hint>The number at the top of your nest marker</mat-hint>
           <mat-error *ngIf="sessionFormGroup.controls['nestControl'].invalid">
-            Invalid nest ID: please enter the ID number on your nest marker
+            Invalid nest ID: your input isn't a number in the range
+            {{MIN_NEST_MARKER}}&ndash;{{MAX_NEST_MARKER}}. Please enter the ID
+            number on your nest marker.
```
floogulinc, line 25: How about
```suggestion
            Invalid nest ID: join codes are numbers between 
            {{MIN_NEST_MARKER}} and {{MAX_NEST_MARKER}}. Please enter the ID
            number on your nest marker.
```

src/app/pages/session-lobby/session-lobby.component.ts
```diff
@@ -51,4 +55,22 @@ export class SessionLobbyComponent implements OnInit {
...
   }
+
+  createJoinCode() {
+    this.joinCodeButtonDisabled$.next(true);
+    this.teacherSessionService.createJoinCode().subscribe(() => {
+      this.joinCodeButtonDisabled$.next(false);
+    }, err => {
+      this.joinCodeButtonDisabled$.next(false);
+      this.matSnackbar.open(
+        `Error: couldn't create a join code. Please try again later.\n(${err})`,
```
floogulinc, line 66: Do we want to include the raw error here?

cypress/integration/join-page.spec.ts
```diff
@@ -130,20 +147,43 @@ describe('The join page', () => {
...
 
       context('Entering bad values into the form', () => {
-        context('Entering a bad session id', () => {
+        describe('Using the malformed join code', () => {
+          for (const badJcId of [
+            'likeZoinksScoob',
+            '1',
+            '1234567',
+            // Eastern Arabic numerals, from Unicode.
+            '٢٢٠٩١٠',
```
floogulinc, line 156: A few more fun tests cases:
```suggestion
            'likeZoinksScoob',
            '1',
            '-123456',
            '-12345',
            '1234567',
            '0123456',
            // Eastern Arabic numerals, from Unicode.
            '٢٢٠٩١٠',
            // Chinese/Japanese numerals
            '一二三四五六',
            // Fullwidth characters
            '１２３４５６'
```

src/app/pages/join-session/join-session.component.html
helloworld12321: Sure—thanks!

src/app/join-code.ts
helloworld12321: Ah, clever! Looks good!

cypress/integration/join-page.spec.ts
helloworld12321: Nice, thanks :)

src/app/pages/session-lobby/session-lobby.component.ts
helloworld12321: We could, but I'd lean towards not troubling our users with that sort of technical stuff ¯\\\_(ツ)\_/¯
Could go either way though!

kklamberty:
To test this, should I be able to use the Vercel deployment or must I run it locally?

src/app/pages/session-lobby/session-lobby.component.ts
floogulinc: Right, it currently *is* printing the error though.

helloworld12321: Oh, sorry :rofl: my bad there :)

kklamberty (approved):
I noticed that I could run into a time when the code should be valid but it isn't valid. I think this is only in cases where it's really soon after a code was deleted and a new code was generated. Perhaps my 'player/participant' browser needed to be refreshed? Other than that, this all looked good. I think it is not likely to come up in practice.

cypress/integration/join-page.spec.ts
```diff
@@ -146,23 +146,66 @@ describe('The join page', () => {
...
+          mockJoinCodeId.replace(/\B(?=([0-9]{3})+(?![0-9]))/g, ' '),
+          // NO-BREAK SPACE
+          mockJoinCodeId.replace(/\B(?=([0-9]{3})+(?![0-9]))/g, '\u00A0'),
+          // CHARACTER TABULATION
+          mockJoinCodeId.replace(/\B(?=([0-9]{3})+(?![0-9]))/g, '\u0009'),
+          // ZERO WIDTH NO-BREAK SPACE
+          mockJoinCodeId.replace(/\B(?=([0-9]{3})+(?![0-9]))/g, '\uFEFF'),
+          mockJoinCodeId.replace(/\B(?=([0-9])+(?![0-9]))/g, ' '),
+          `   ${mockJoinCodeId.replace(/\B(?=([0-9]{2})+(?![0-9]))/g, '   ')}   `,
+        ]) {
```
helloworld12321, line 160: Those are quite the regular expressions! :joy: I appreciate the fanciness, but do you think we could do something like …? (It's more boring, to be sure, but this way I don't have to remember what a negative lookbehind does :stuck_out_tongue_winking_eye:)

```suggestion
        for (const goodJcId of [
          mockJoinCodeId.match(/[0-9]{3}/g).join(' '),
          // NO-BREAK SPACE
          mockJoinCodeId.match(/[0-9]{3}/g).join('\u00A0'),
          // CHARACTER TABULATION
          mockJoinCodeId.match(/[0-9]{3}/g).join('\u0009'),
          // ZERO WIDTH NO-BREAK SPACE
          mockJoinCodeId.match(/[0-9]{3}/g).join('\uFEFF'),
          mockJoinCodeId.split('').join(' '),
          `   ${mockJoinCodeId.match(/[0-9]{2}/g).join('   ')}   `,
        ]) {
```

helloworld12321 (approved):

helloworld12321:
@kklamberty Oh, funny—are you saying that a student couldn't join the session using the new code?

UtkarshKr007 (approved):
Couldn't get any undesired results past the join code with the new validation!! Perfect.
The codes did expire for me after an hour, so that's good. Great work on optimazations and refactoring the firestore tests.

kklamberty (approved):
I used my phone and computer as participants and hosted on my computer. I tried using join codes that had been removed (failed as it should), different codes for each participant (removing code in between), and using a join code for the same session more than one time after closing the participant view and using the back button. It seems to work on all counts, and I think the interface looks great. It allows you to enter your name when you join again, so even without renaming working you can rename yourself if you do it wrong the first time. Nice work!

PR #173, opened 2020-12-23
Implement renaming and removing students

floogulinc:
image

image
image
image

image
image


audreylemeur (approved):

kklamberty (changes requested):
Removing a student removes their name from the lobby, but when the round begins, they are still very much a part of it from what I can see. It's interesting, because they are not assigned a bee... but they still have an AR view. I propose that we need to look into what happens from the student side of things. I would be okay with approving this aspect and creating a new issue for removing the student from the session and handling what happens to the student's view in that case, but I don't think this should move ahead without something in place for that.

floogulinc:
> Removing a student removes their name from the lobby, but when the round begins, they are still very much a part of it from what I can see. It's interesting, because they are not assigned a bee... but they still have an AR view. I propose that we need to look into what happens from the student side of things. I would be okay with approving this aspect and creating a new issue for removing the student from the session and handling what happens to the student's view in that case, but I don't think this should move ahead without something in place for that.

@kklamberty Joe and I have tested this and realized that it's not actually an issue. Students who are removed from a session will always sit at the "Waiting for round to begin" screen. The reason you were able to see the round start was because you were testing the teacher and student side in the same browser session so both of them were actually the same user ID, so even after removing the student it could still read the round because it was actually the teacher as well.

floogulinc:
@kklamberty @audreylemeur We have just added a dialog that appears on the student side when they are removed from a session. Also we fixed the student display's "invalids session" message when trying to load a session that either doesn't exist or you don't have access to.

kklamberty (approved):
Thanks for testing and helping me understand how that odd case occurred. That makes sense. 

I like the dialog for when you've been removed from a session, and I think this PR looks great!

UtkarshKr007 (approved):
I like the new dialog informing student they left the round and providing them an option to go back to the home page. Looks great.

PR #174, opened 2020-12-23
Implement active periods for bees

### LINKED ISSUE #161: Implement active periods for bees
Different bees are active at different times. We need to make it obvious when a bee is able to do things and when it's not yet active or no longer active.

floogulinc:
Closes #161

image

We can definitely do some work on the UI that's shown to students while their bee is inactive.


audreylemeur (approved):
Looks good. The one thing to note is that for me, the camera permissions appear only after the bee is active.

floogulinc:
> Looks good. The one thing to note is that for me, the camera permissions appear only after the bee is active.

Good point. We should be asking for camera perms earlier in the flow but that's probably something for a separate PR. Thanks.

kklamberty (approved):
Sometimes, the spacing of the message about your bee being inactive looks a little squished (for long bee names), but I don't think it's a dealbreaker by any means. Looks good!

PR #178, opened 2020-12-26
Use self-hosted fonts

floogulinc:


src/app/pages/session-lobby/session-lobby.component.scss
```diff
@@ -31,7 +31,7 @@
   mat-card-actions {
     display: flex;
     justify-content: space-between;
-    align-items: flex-end;
+    align-items: center;
```
helloworld12321, line 34: Ah, cool. Nice catch.

helloworld12321 (approved):
This is very nice—thanks, Paul!

Looks like like the service worker is indeed picking up the fonts: 🙂

Screen Shot 2020-12-28 at 10 14 54 PM

UtkarshKr007 (approved):

PR #182, opened 2020-12-31
8 flower rounds

floogulinc:
We now allow round templates to have either 8 or 16 flowers and have added an example 8 flower round.


audreylemeur (approved):
Great job. For a later pull request, we might want to give the option of generating 8 flower barcodes instead of 16.

kklamberty (approved):
This seems to work just fine for me. Thanks for finishing up the work on this! Much appreciated.

PR #183, opened 2020-12-31
Prefer markers near the center of the screen

floogulinc:
This makes it so when deciding which marker to display the name of and interact button for when there are multiple visible it will prefer markers that are near the center of the screen and closer.


audreylemeur (approved):
The feature works well even though I seem to be having issues with the pollen counter.

helloworld12321:
@audreylemeur Oh, sorry to hear that. What sort of issues?

kklamberty (approved):
I think this works very nicely. Good work!

PR #184, opened 2021-01-02
Update THREEAR to use official repo

floogulinc:
Currently the update to `v1.0.0` isn't published on NPM yet but if it is we should switch to that rather than using git. I have created an issue in the THREEAR repo about this: https://github.com/JamesLMilner/THREEAR/issues/116.


floogulinc:
This is failing because it makes the main bundle too large (over 5mb) so it will pretty much require code splitting with modules (#185).

floogulinc:
> Looks good! I'm glad they fixed our bug. 🙂
> Should we archive the mn-pollinators/THREEAR repo after this gets merged?

Yep! That's the plan.

kklamberty (approved):
The AR view for the students still seems to work with this change. I don't notice any differences, and my impression is that no news is good news on that front. Thanks!

PR #186, opened 2021-01-03
Add robots.txt, description meta tag, and some image alt tags

floogulinc:
A `robots.txt` file is needed for search engines to be able to index the site.

### DIFF
--- a/src/app/components/hill-background/hill-background.component.html
+++ b/src/app/components/hill-background/hill-background.component.html
@@ -1,1 +1,1 @@
-<img src="/assets/art/other/hill-1000w.jpg" srcset="/assets/art/other/hill-1000w.jpg 1000w, /assets/art/other/hill-2000w.jpg 2000w, /assets/art/other/hill-4000w.jpg 4000w" sizes="100vw"/>
+<img src="/assets/art/other/hill-1000w.jpg" srcset="/assets/art/other/hill-1000w.jpg 1000w, /assets/art/other/hill-2000w.jpg 2000w, /assets/art/other/hill-4000w.jpg 4000w" sizes="100vw" alt="An illustration of a green, grassy hill"/>
--- a/src/app/components/hill-background/hill-background.component.html
+++ b/src/app/components/hill-background/hill-background.component.html
@@ -1,1 +1,1 @@
-<img src="/assets/art/other/hill-1000w.jpg" srcset="/assets/art/other/hill-1000w.jpg 1000w, /assets/art/other/hill-2000w.jpg 2000w, /assets/art/other/hill-4000w.jpg 4000w" sizes="100vw" alt="An illustration of a green, grassy hill"/>
+<img src="/assets/art/other/hill-1000w.jpg" srcset="/assets/art/other/hill-1000w.jpg 1000w, /assets/art/other/hill-2000w.jpg 2000w, /assets/art/other/hill-4000w.jpg 4000w" sizes="100vw" alt=""/>
--- a/src/app/pages/about/about.component.html
+++ b/src/app/pages/about/about.component.html
@@ -51,7 +51,7 @@
       <mat-card-content>
         <mat-nav-list>
           <mat-list-item *ngFor="let contributor of allContributors">
-            <img matListAvatar src="{{contributor.avatarUrl}}&size=40" srcset="{{contributor.avatarUrl}}&size=80 2x, {{contributor.avatarUrl}}&size=160 4x" alt="{{contributor.username}} GitHub avatar">
+            <img matListAvatar src="{{contributor.avatarUrl}}&size=40" srcset="{{contributor.avatarUrl}}&size=80 2x, {{contributor.avatarUrl}}&size=160 4x" alt="@{{contributor.username}}">
             <a matLine [href]="contributor.url">{{contributor.name || contributor.username}}</a>
           </mat-list-item>
         </mat-nav-list>
--- a/src/app/pages/about/about.component.html
+++ b/src/app/pages/about/about.component.html
@@ -6,7 +6,7 @@
     </button>
     <mat-card>
       <mat-card-header>
-        <img mat-card-avatar src="/assets/icons/icon-circle.svg">
+        <img mat-card-avatar src="/assets/icons/icon-circle.svg" alt="Buzz About icon">
         <mat-card-title>Buzz About</mat-card-title>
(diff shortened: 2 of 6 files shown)


src/app/pages/about/about.component.html
```diff
@@ -6,7 +6,7 @@
     </button>
     <mat-card>
       <mat-card-header>
-        <img mat-card-avatar src="/assets/icons/icon-circle.svg">
+        <img mat-card-avatar src="/assets/icons/icon-circle.svg" alt="Buzz About icon">
```
helloworld12321, line 9:
```suggestion
        <img mat-card-avatar src="/assets/icons/icon-circle.svg" alt="">
```

src/app/pages/about/about.component.html
```diff
@@ -51,7 +51,7 @@
       <mat-card-content>
         <mat-nav-list>
           <mat-list-item *ngFor="let contributor of allContributors">
-            <img matListAvatar src="{{contributor.avatarUrl}}&size=40" srcset="{{contributor.avatarUrl}}&size=80 2x, {{contributor.avatarUrl}}&size=160 4x">
+            <img matListAvatar src="{{contributor.avatarUrl}}&size=40" srcset="{{contributor.avatarUrl}}&size=80 2x, {{contributor.avatarUrl}}&size=160 4x" alt="{{contributor.username}} GitHub avatar">
```
helloworld12321, line 54:
```suggestion
            <img matListAvatar src="{{contributor.avatarUrl}}&size=40" srcset="{{contributor.avatarUrl}}&size=80 2x, {{contributor.avatarUrl}}&size=160 4x" alt="">
```

src/app/components/hill-background/hill-background.component.html
```diff
@@ -1 +1 @@
-<img src="/assets/art/other/hill-1000w.jpg" srcset="/assets/art/other/hill-1000w.jpg 1000w, /assets/art/other/hill-2000w.jpg 2000w, /assets/art/other/hill-4000w.jpg 4000w" sizes="100vw"/>
+<img src="/assets/art/other/hill-1000w.jpg" srcset="/assets/art/other/hill-1000w.jpg 1000w, /assets/art/other/hill-2000w.jpg 2000w, /assets/art/other/hill-4000w.jpg 4000w" sizes="100vw" alt="An illustration of a green, grassy hill"/>
```
helloworld12321, line 1:
```suggestion
<img src="/assets/art/other/hill-1000w.jpg" srcset="/assets/art/other/hill-1000w.jpg 1000w, /assets/art/other/hill-2000w.jpg 2000w, /assets/art/other/hill-4000w.jpg 4000w" sizes="100vw" alt=""/>
```

src/app/pages/about/about.component.html
```diff
@@ -63,7 +63,7 @@
       <mat-card-title>Funding</mat-card-title>
       <mat-card-content>
         <div class="enrtf-logo-container">
-          <img src="/assets/enrtf-logo.png">
+          <img src="/assets/enrtf-logo.png" alt="ENRTF logo">
```
helloworld12321, line 66:
```suggestion
          <img src="/assets/enrtf-logo.png" alt="">
```

src/app/pages/home/home.component.html
```diff
@@ -2,7 +2,7 @@
 <div class="home-container">
   <div class="main-section">
     <mat-card class="column">
-      <img mat-card-image src="/assets/icons/icon-square.svg">
+      <img mat-card-image src="/assets/icons/icon-square.svg" alt="Buzz About icon">
```
helloworld12321, line 5:
```suggestion
      <img mat-card-image src="/assets/icons/icon-square.svg" alt="">
```

helloworld12321 (changes requested):
I really like the `meta name="Description"` that you wrote for us!

My only real question is whether we might want to use empty `alt` attributes (`alt=""`) for these images. It looks like the Web Content Accessibility Guidelines recommend that for images that are either purely decorative or redundant with nearby text descriptions: https://www.w3.org/WAI/tutorials/images/decorative/

(If you don't want to do that, though, I'm fine to leave things as they are! 🙂 Everything looks good.)

helloworld12321:
(It seems like Google would rather we use `<meta name="robots" content="noindex"/>` directives instead of putting the pages in the robots.txt

(https://developers.google.com/search/docs/advanced/robots/intro#what-is-robots.txt-used-for)

but I'm not particularly concerned about that, especially given all of the extra complexity we'd need to add those `meta` tags. The robots.txt seems like it'll work just fine for our needs.)

src/app/pages/about/about.component.html
floogulinc: Since this is a clickable element in a list I think it should have an alt tag, even GitHub themselves give your avatar on this page `alt="@helloworld12321"`.

What do you think about matching their behavior and doing this: 
```suggestion
            <img matListAvatar src="{{contributor.avatarUrl}}&size=40" srcset="{{contributor.avatarUrl}}&size=80 2x, {{contributor.avatarUrl}}&size=160 4x" alt="@{{contributor.username}}">
```

helloworld12321:
Also, you asked a while ago about why we would want to specify `alt=""` instead of just omitting the alt attribute entirely. My impression is that if you omit the alt attribute, non-visual browsers like Lynx might try to replace the image with its filename. If you set `alt=""`, though, they'll just omit it.

Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Attributes

floogulinc:
> (It seems like Google would rather we use `<meta name="robots" content="noindex"/>` directives instead of putting the pages in the robots.txt
> 
> [developers.google.com/search/docs/advanced/robots/intro#what-is-robots.txt-used-for](https://developers.google.com/search/docs/advanced/robots/intro#what-is-robots.txt-used-for)
> 
> but I'm not particularly concerned about that, especially given all of the extra complexity we'd need to add those `meta` tags. The robots.txt seems like it'll work just fine for our needs.)

Well, with Angular, every page is `index.html` so we can't have different meta tags for specific pages.

helloworld12321:
Ooh, I didn't know that about Vercel—cool!

helloworld12321 (approved):
Looks excellent&mdash;thanks, Paul! 😃

helloworld12321 (approved):

helloworld12321 (approved):

UtkarshKr007 (approved):
Looks great. Nice work

PR #187, opened 2021-01-03
Disable pointer events on hill background

floogulinc:
This disables pointer events on the hill background which prevents accidently dragging it around and such. It makes it much more like a regular background set with CSS.


UtkarshKr007 (approved):
Tested it.

kklamberty (approved):

PR #190, opened 2021-01-04
Create back-button component

### LINKED ISSUE #105: Back/exit button for student display

floogulinc:
Closes #105


audreylemeur (approved):

src/app/components/back-button/back-button.component.scss
```diff
@@ -0,0 +1,9 @@
+:host[top-left] {
```
helloworld12321, line 1: Ooh, that's a nice way of doing things.

src/app/pages/home/home.component.html
```diff
@@ -2,7 +2,7 @@
 <div class="home-container">
   <div class="main-section">
     <mat-card class="column">
-      <img mat-card-image src="/assets/icons/icon-square.svg">
+      <img width="1" height="1" mat-card-image src="/assets/icons/icon-square.svg">
```
helloworld12321, line 5: Do you think it would be all right if you added a comment explaining what the `width` and `height` attributes do?

(I'm having trouble figuring out what's going on here. Googling suggests that this might be a trick used to get an image into the browser cache extra-quickly? Am I getting that right?)

Thanks a bunch! 🙂

src/app/pages/home/home.component.html
```diff
@@ -26,11 +26,24 @@
...
+          </button>
+          <mat-menu #menu="matMenu">
+            <button mat-menu-item routerLink="/markers">
+              Markers
+            </button>
+            <button mat-menu-item routerLink="/about">
+              About
+            </button>
+          </mat-menu>
+        </span>
```
helloworld12321, line 46: Ooh, that looks quite nice.

src/app/pages/student-display/student-display.component.html
```diff
@@ -1,35 +1,35 @@
...
-      <mat-card>
+      <mat-card *ngSwitchDefault>
         <mat-card-title>
           Loading Session&hellip;
         </mat-card-title>
         <mat-spinner></mat-spinner>
       </mat-card>
+
     </div>
   </ng-container>
```
helloworld12321, line 35: Ah, okay, I see what you did here.

I'd prefer to avoid nested `ngSwitch`s if it's possible, though—HTML tends to be very deeply indented, and it can be difficult to identify whether a given case belongs to the inner or outer switch. I know we would have to write the `<app-hill-background cover></app-hill-background>` element once for each NgSwitchCase, but I think the extra readability would be worth it.

If you disagree, though, I'd also be okay with leaving it as it is :)

src/app/pages/student-round/student-round.component.html
```diff
@@ -1,23 +1,19 @@
...
-    <div class="centered-container">
-      <mat-card>
+      <mat-card *ngSwitchCase="ScreenId.InactiveBee">
         <mat-card-title>Your {{(roundService.currentBeeSpecies$ | async).name}} is currently inactive</mat-card-title>
       </mat-card>
+
     </div>
-	</ng-container>
+  </ng-container>
 </ng-container>
```
helloworld12321, line 19: Same comment here.

---

Incidentally, I do like how you saved us an extra `ng-container` by going from

```html
  <ng-container *ngSwitchCase="ScreenId.Play">
     <app-play-round></app-play-round>
   </ng-container>
```

to

```html
   <app-play-round *ngSwitchCase="ScreenId.Play"></app-play-round>
```

helloworld12321 (changes requested):
Okay, I'm requesting two changes here, one small and one medium-sized. (As always, feel free to disagree with these suggestions. If you prefer how it is right now, I'm fine with that too. 🙂)

The small suggestion is to add a comment on the home screen about the how `width="1" height="1"` works. (I tried to google it, but I had trouble. Is it some sort of caching trick?)

The medium suggestion is to flatten out the `ngSwitch`s in the StudentDisplayComponent and StudentRoundComponent. I see what you're going for here—it does save us from repeating the `app-hill-background` once for each case—but nested switches are hard to wrap your head around 🙃

The UI does look very nice, though—I like your aesthetic choices!

Okay, that's all I have. Thanks! ☺️

src/app/pages/student-display/student-display.component.html
```diff
@@ -1,35 +1,35 @@
...
-      <mat-card>
+    <div class="centered-container" [ngSwitch]="currentScreen$ | async">
+
+      <mat-card *ngSwitchCase="ScreenId.SessionLobby">
         <mat-card-title>Waiting for round to begin&hellip;</mat-card-title>
+        <mat-card-content>
+          Please wait for the host to start a round.
+        </mat-card-content>
+        <mat-card-actions align="end">
+          <button mat-button routerLink="/">Exit Session</button>
```
helloworld12321, line 15: And then, just checking to make sure—we don't want this button to remove the student from the session in Firebase, do we?

src/app/pages/home/home.component.html
floogulinc: I'm doing it to hint to the browser what the aspect ratio of the image is (1x1 is arbitrary, any x/y=1 would work) to prevent layout shift when the image loads.

src/app/pages/student-display/student-display.component.html
floogulinc: We don't currently have that implemented so I think that would be best left for another PR in the future if it is something we want.

Currently the component calls `sessionService.leaveSession()` when destroyed so if in the future we wanted that to be the case we could implement it there maybe.

src/app/pages/student-display/student-display.component.html
floogulinc: I was noticing that as a student since these screens tend to be switched out quite a bit it really didn't look good and was distracting when it flickered white everytime something changed. Even with this change there is still a bit of flicker that can happen when transitioning between session lobby and round. Honestly we may want to refactor the entire student display to not use the nested `student-round` component.

src/app/pages/home/home.component.html
floogulinc: I'm not sure I want to add a comment about it into the HTML since HTML comments don't get stripped away when the app is compiled.

helloworld12321: Ah, I see, that makes sense! Cool!

Hmm… I can see why you might feel that way about HTML comments. What about a comment in `home.component.scss` instead? It's a bit out of the way, but we are adjusting the size of the mat-card-image there, so that might be a good place to mention it :)

src/app/pages/student-display/student-display.component.html
helloworld12321: Oh, okay, if there's a practical benefit, then I'm all for doing things this way. Thanks for explaining that! 🙂

src/app/pages/home/home.component.html
floogulinc: Great idea. Done!

helloworld12321 (approved):
Looks great—thanks very much!

helloworld12321 (dismissed):

kklamberty (approved):

kklamberty (approved):

helloworld12321 (approved):
Looks good! Thanks :)

PR #192, opened 2021-01-08
Create round monitor page

floogulinc:
image


helloworld12321:
We're keeping this branch around, but we're probably not going to merge it into master. (We'll be using it later, though!) 🙂

floogulinc:
This is being integrated into #209.

PR #193, opened 2021-01-09
Import 'firebase/auth' in auth.service.ts

floogulinc:
This fixes the auth issues we were having during play testing by forcing it to bundle `firebase/auth` into the main bundle so it is always loaded when we need it for firestore operations.

### DIFF
--- a/src/app/services/auth.service.ts
+++ b/src/app/services/auth.service.ts
@@ -4,7 +4,7 @@ import { switchMap, map, shareReplay } from 'rxjs/operators';
 import { from, of, Observable } from 'rxjs';
 
 // Make sure that firebase/auth is included in the main bundle
-// (not lazy-loaded).
+// (not lazy-loaded, because relying on it to lazy-load was not working)
 import 'firebase/auth';
 
 @Injectable({
--- a/src/app/services/auth.service.ts
+++ b/src/app/services/auth.service.ts
@@ -2,6 +2,9 @@ import { Injectable } from '@angular/core';
 import { AngularFireAuth } from '@angular/fire/auth';
 import { switchMap, map, shareReplay } from 'rxjs/operators';
 import { from, of, Observable } from 'rxjs';
+
+// Make sure that firebase/auth is included in the main bundle
+// (not lazy-loaded).
 import 'firebase/auth';
 
 @Injectable({
--- a/src/app/services/auth.service.ts
+++ b/src/app/services/auth.service.ts
@@ -1,8 +1,8 @@
 import { Injectable } from '@angular/core';
 import { AngularFireAuth } from '@angular/fire/auth';
-import { FirebaseService } from './firebase.service';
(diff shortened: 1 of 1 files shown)


helloworld12321:
Oh, thank you so much! Sorry for not catching that during code review for #185 :(

kklamberty (dismissed):

helloworld12321 (approved):
Looks good! As near as I can tell, that fixed everything :)

Thanks very much for diagnosing & fixing that.

I'm not sure why the auth module wasn't being loaded previously—some quirk of `@angular/fire/auth`? (This lazy-loading stuff is all pretty new to me. 🙂)

---

In the future, we might want to look into marking more explicitly what actions trigger the anonymous login. (I'm imagining either route guards, or some sort of wrapper function like

```ts
function requireLogin<T>(obs$: Observable<T>): Observable<T>
```

but I'm not sure whether that would work or not. 🙂)

From what I can tell, though, right now you get an anonymous user wherever you import `StudentSessionService` or `TeacherSessionService`, which seems like a fine thing! :)

src/app/services/auth.service.ts
```diff
@@ -1,16 +1,16 @@
 import { Injectable } from '@angular/core';
 import { AngularFireAuth } from '@angular/fire/auth';
-import { FirebaseService } from './firebase.service';
 import { switchMap, map, shareReplay } from 'rxjs/operators';
 import { from, of, Observable } from 'rxjs';
+import 'firebase/auth';
```
helloworld12321, line 5: Oh, I know I just approved this PR, but it might make sense to comment this import.

```suggestion

// Make sure the auth service is included in the main bundle
// (not lazy-loaded).
import 'firebase/auth';
```

helloworld12321 (dismissed):

src/app/services/auth.service.ts
```diff
@@ -1,16 +1,19 @@
 import { Injectable } from '@angular/core';
 import { AngularFireAuth } from '@angular/fire/auth';
-import { FirebaseService } from './firebase.service';
 import { switchMap, map, shareReplay } from 'rxjs/operators';
 import { from, of, Observable } from 'rxjs';
 
+// Make sure that firebase/auth is included in the main bundle
+// (not lazy-loaded).
```
kklamberty, line 7:
```suggestion
// (not lazy-loaded, because relying on it to lazy-load was not working). 
```

kklamberty (dismissed):
I suggested a change, but I'm indifferent about its inclusion. I find the comment as written just a little cryptic. My understanding of what was happening may be incorrect. Was it lazy-loading before, and that was just not quite working how we wanted so we need to be sure to explicitly import it here so it's available in time?

src/app/services/auth.service.ts
helloworld12321: Oh, fair ☺️ Sorry, my understanding of the issue is also shaky, so I didn't write a very detailed comment :)

PR #194, opened 2021-01-09
Avoid sequential nest interactions from students

UtkarshKr007:
As we saw during the playtesting, sometimes there are multiple nest interactions with only the initial interaction adding anything meaningful to the simulation. We can avoid this by 

- disabling the intearction option with a nest barcode when the bee's pollen count is at 0. 
- Having a note like "Gather pollen to deposit them in your nest"

The reason I think we should disable the button is any visual or textual feedback might guide users but won't ultimately guarantee avoiding the behavior. Also when low on a time, it is expected students to be hurriedly tapping the button as they get to their nest, so any sort of feedback would just be another layer that would get repeated with every extra nest Interaction.

The one reason I can think against disabling the button is if in our research we want to observe how easy it for students to follow textual or visual instructions. Perhaps even compare whether textual or visual feedback is more effective in stopping extra interactions.


src/app/components/play-round/play-round.component.scss
```diff
@@ -77,6 +76,14 @@
   text-shadow: 0 0 0.6em #111;
 }
 
+.marker-name {
+  @include responsive-font-size(2rem, 2.5rem, 3rem);
+}
+
+.nest_guide {
+  @include responsive-font-size(1.1rem, 1.5rem, 2rem);
```
floogulinc, line 84: Any particular reason this is `1.1rem` rather than `1rem`? Also, we usually don't use underscores for class names.

src/app/components/play-round/play-round.component.html
```diff
@@ -23,8 +23,8 @@
   <div class="marker-name">
     {{marker.name | titlecase}}
   </div>
-  <div class="nest-guide" *ngIf="marker.isNest && !marker.canVisit">
-    Gather Pollen to deposit them in your nest
+  <div class="marker-tip" *ngIf="showTip(marker) as tip">
```
floogulinc, line 26: Don't use functions in data binding. The tip and logic for creating it should be moved to the observables for the markers.

floogulinc (changes requested):

UtkarshKr007:
We can discuss whether we want to go with this approach during Monday's meeting. Until then, doesn't make much sense to ask people to review this throughly.

src/app/components/play-round/play-round.component.scss
```diff
@@ -60,14 +60,13 @@
   }
 }
 
-.marker-name {
+.marker-name, .marker-tip {
```
floogulinc, line 63: Since you put both the `marker-name` and `marker-tip` divs in a wrapper container, shouldn't we do the positioning and some margins for it on the container?

src/app/components/play-round/play-round.component.ts
```diff
@@ -113,4 +118,11 @@ export class PlayRoundComponent implements OnInit {
     // Normalize scale
     return ((scale - 1) * 0.2) + 1;
   }
+
+  showTip(marker: RoundMarker) {
+    if (marker.isNest && !marker.canVisit) {
+      marker.tip = 'Gather Pollen to deposit them in your nest';
+    }
+  }
```
floogulinc, line 126: I'm not a fan of functions that mutate their arguments. Could we have something like getTip(marker) that returns a string of the right tip?

src/app/markers.ts
```diff
@@ -31,6 +31,9 @@ export interface RoundMarker extends ARMarker {
   isNest: boolean;
 
   canVisit: boolean;
```
floogulinc, line 33:
```suggestion
  canVisit: boolean;
  
```

src/app/components/play-round/play-round.component.ts
```diff
@@ -34,15 +34,17 @@ export class PlayRoundComponent implements OnInit {
...
-    map(([student, bee]) => ({
+    map(([student, bee, pollenCount]) => ({
       name: bee.nest_type.name,
       isNest: true,
-      canVisit: true,
+      canVisit: pollenCount !== 0,
       barcodeValue: student.nestBarcode,
-      imgPath: `/assets/art/512-square/nests/${bee.nest_type.art_file}`
+      imgPath: `/assets/art/512-square/nests/${bee.nest_type.art_file}`,
+      tip: pollenCount === 0 ? 'Gather Pollen to deposit them in your nest' : null
```
floogulinc, line 47:
```suggestion
      tip: pollenCount === 0 ? 'Gather pollen to deposit in your nest' : null
```

floogulinc (changes requested):

src/app/components/play-round/play-round.component.scss
```diff
@@ -77,6 +80,14 @@
   text-shadow: 0 0 0.6em #111;
 }
 
+.marker-name {
+  @include responsive-font-size(2rem, 2.5rem, 3rem);
+}
+
+.marker-tip {
+  @include responsive-font-size(1rem, 1.5rem, 2rem);
```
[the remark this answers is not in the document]
UtkarshKr007: Good catch

src/app/components/play-round/play-round.component.spec.ts
```diff
@@ -226,7 +226,7 @@ describe('PlayRoundComponent', () => {
           ['name', allBeeSpecies.apis_mellifera.nest_type.name],
           ['isNest', true],
           ['barcodeValue', 30],
-          ['canVisit', true],
+          ['canVisit', mockBeePollen$.value !== 0],
```
helloworld12321, line 229: Oh, cool, I didn't know that `BehaviorSubject.value` was a thing!

src/app/components/play-round/play-round.component.spec.ts
```diff
@@ -321,5 +321,29 @@ describe('PlayRoundComponent', () => {
...
+      mockBeePollen$.next(0);
+      mockBeePollen$.next(1);
+      mockBeePollen$.next(0);
+      tick(0);
+
+      expect(emittedNestMarkers.length).toBe(3);
+      expect(emittedNestMarkers[0].canVisit).toBe(false);
+      expect(emittedNestMarkers[1].canVisit).toBe(true);
+      expect(emittedNestMarkers[2].canVisit).toBe(false);
+    }));
```
helloworld12321, line 347: This is a really nice test. Thanks for writing that, Utkarsh!

helloworld12321 (approved):
Works like a charm—thanks, Utkarsh!

src/app/components/play-round/play-round.component.scss
```diff
@@ -60,14 +60,17 @@
...
   margin-bottom: 0;
   margin-right: 8px;
   margin-left: 8px;
+}
+
+.marker-name, .marker-tip {
   text-align: center;
 
-  @include responsive-font-size(2rem, 2.5rem, 3rem);
   font-weight: bold;
```
floogulinc, line 74: Sorry for this nitpicking but I think we should use a lower font-weight. bold (which is 700) looks like a bit much, especially on smaller screens and with the smaller tip. Using this font weight also requires  the browser to get another font file.

```suggestion
  font-weight: 500;
```

floogulinc (changes requested):

src/app/components/play-round/play-round.component.scss
UtkarshKr007: Sure. I don't have a preference towards either.

floogulinc (approved):

helloworld12321 (approved):

Issue #131, opened 2020-08-14
Get camera permissions ahead of time


floogulinc:
We should ask students for camera permissions ahead of using AR to streamline the flow. I think a good point would be to ask for it upon entering the session if its not already granted. We may want to add a custom dialog for this to explain and also properly handle the case where it is denied. 

This comes from @kklamberty's suggestion in https://github.com/mn-pollinators/buzz-about/pull/130#pullrequestreview-467799971.

helloworld12321:
Given that

1. There's a lot of cross-browser variation in this issue, so
2. Solving this might introduce more complexity/flakiness than we have right now, and
3. I think we're at least somewhat satisfied with where this is right now,

I'm going to close this issue as `wontfix`.

(Should we ever want to consider this issue again, we can always open it back up. 🙂)

PR #199, opened 2021-01-12
Bug fixes and refactoring in the AR View component

floogulinc:
This PR cleans up and attempts to fix some bugs in the AR View component. Testing for all of these things is fairly difficult so it would be great to test this on real devices to see if some of the issues we noticed in play testing are gone.

Specifically, we hope to have fixed:
- The issues with just getting a white screen sometimes instead of the AR view
- AR view just totally freezing on Android

This PR also removes the old debug view for AR which we haven't used in a while.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12198,11 +12198,6 @@
         "wrappy": "1"
       }
     },
-    "info-monitor": {
-      "version": "1.0.12",
-      "resolved": "https://registry.npmjs.org/info-monitor/-/info-monitor-1.0.12.tgz",
-      "integrity": "sha512-1/Em3oSZ8z4Tppj55ihlQh8YMUdDmrjZTL1hsNUFzyLt17KpOGR7p/zPI1XP1LKZ8Zlca7CfiZ8lVSjRVexaug=="
-    },
     "inherits": {
       "version": "2.0.4",
       "resolved": "https://registry.npmjs.org/inherits/-/inherits-2.0.4.tgz",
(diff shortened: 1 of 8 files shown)


helloworld12321:
(I probably shouldn't review this given that I helped work on it.
Granted, not a lot of it is my code, but I do think a new pair of eyes would be helpful 🙂)

UtkarshKr007:
We might have to look at issue: #131 before solving the white screen problem.
I can consistently reproduce that issue by having the teacher pause and play the round before the student has given the camera permission. This will keep the screen white even after the permission has been granted.

These are the errors in the console: 


Notably, the second error regarding 'parentNode of undefined' occurs after the what's supposed to be ar view has ended.

UtkarshKr007 (commented):
Should we also delete the ar-view-component spec file since it's empty.

floogulinc:
> We might have to look at issue: #131 before solving the white screen problem.
> I can consistently reproduce that issue by having the teacher pause and play the round before the student has given the camera permission. This will keep the screen white even after the permission has been granted.

Thanks for catching this, I'll take a look. I have been thinking of looking at using the new browser permissions APIs to try and request perms as soon as the session starts (at least on supported browsers, aka not Safari). I might try and do that in this PR.

floogulinc:
@UtkarshKr007 In my testing what I see happen in that case is as long as the student did hit the allow button, it will work after another pause/play cycle.

floogulinc:
After more experimentation, truly fixing the permissions issues is going to be a major thing so I'm reopening this for review.

UtkarshKr007 (approved):

UtkarshKr007 (approved):

kklamberty (approved):
@UtkarshKr007 I appreciate your comment about how to recreate the white screen issue. Thanks! I still do have that white screen issue, but the permissions for the camera on subsequent rounds do seem to stick once I've given them. I don't have an Android device to test that bug. Thanks for your work on this, Joe and Paul!

PR #200, opened 2021-01-13
Add feedback and animations for collecting pollen

floogulinc:


floogulinc:
Here's a short screen recording of the animations in action:

https://user-images.githubusercontent.com/1300395/105741186-a1411300-5eff-11eb-9d5e-f99072dda70d.mp4

audreylemeur (changes requested):
I don't know what exactly is wrong, but when I click create join code, it flashes the code for less than a second then the button reappears.

floogulinc:
> I don't know what exactly is wrong, but when I click create join code, it flashes the code for less than a second then the button reappears.

@audreylemeur I'm not having this issue on the Vercel deployment. Are you running it locally with the Firebase emulator?

helloworld12321:
@audreylemeur Oh, weird! Can I ask two more questions about that?

1. Do you get any errors in the browser console?
2. Would you be able to make a screen recording to show a) the circumstances under which you're finding this bug, and b) what the bug looks like when it happens?

Thanks very much for the bug report!

audreylemeur:
> > I don't know what exactly is wrong, but when I click create join code, it flashes the code for less than a second then the button reappears.
> 
> @audreylemeur I'm not having this issue on the Vercel deployment. Are you running it locally with the Firebase emulator?
> 
> This may be caused by your computer's clock being incorrect, try making sure it's syncing with an NTP server.

Thanks Paul that's what the issue was. I've dual booted my computer and sometimes it doesn't sync the clock immediately when I switch os.

audreylemeur (commented):
Maybe this was already discussed but did we want this text to be all lowercase:
image
The animation looks really good otherwise.

floogulinc:
> Thanks Paul that's what the issue was. I've dual booted my computer and sometimes it doesn't sync the clock immediately when I switch os.

@audreylemeur Ah yes that totally makes sense since Linux by defaults expects the real-time clock to use UTC while Windows sets it to your local timezone. You can actually [set Windows to use UTC as well](https://wiki.archlinux.org/index.php/System_time#UTC_in_Windows) to prevent this issue from happening when you switch OSes.

By the way, was this happening when testing locally with the emulator or was it happening with the Vercel deployment?

floogulinc:
> Maybe this was already discussed but did we want this text to be all lowercase:
> image

We have issue #217 open to address that.

audreylemeur (approved):

audreylemeur:
> > Thanks Paul that's what the issue was. I've dual booted my computer and sometimes it doesn't sync the clock immediately when I switch os.
> 
> @audreylemeur Ah yes that totally makes sense since Linux by defaults expects the real-time clock to use UTC while Windows sets it to your local timezone. You can actually [set Windows to use UTC as well](https://wiki.archlinux.org/index.php/System_time#UTC_in_Windows) to prevent this issue from happening when you switch OSes.
> 
> By the way, was this happening when testing locally with the emulator or was it happening with the Vercel deployment?

Vercel deployment

kklamberty (approved):
These animations are very engaging! I think it looks great.

UtkarshKr007 (approved):
The animation for collection pollen looks great.

PR #201, opened 2021-01-13
Scale up some of our base font sizes and prevent pointer events and selections in play-round

floogulinc:


helloworld12321:
The main theme of this pull request is that we want to minimize the amount of zooming that happens in AR.

Part of this is achieved by disabling touchscreen-gestures on the AR view.

But the font size changes are just as important; iOS Safari has an "automatically zoom in on form fields" feature that it enables when the font is smaller than 16px. ([source](https://stackoverflow.com/a/6394497))

Right now, we've disabled all touchscreen zoom-y viewport-size-change-y gestures on the AR view, which looks something like this:

```scss
:host {
  …
  touch-action: none;
}
```

However, we're open to leaving the core pinch-to-zoom gesture enabled, while disabling secondary gestures like double-tap-to-zoom. That would look something like this:

```scss
:host {
  …
  touch-action: manipulation;
}
```

For reference, here's the MDN documentation for CSS `touch-action`:  
<https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action>

helloworld12321:
(The other theme of this pull request is to make the text and images in AR non-selectable.)

PR #202, opened 2021-01-14
Simplify the pollen indicator ngFor and make MAX_CURRENT_POLLEN general

floogulinc:
This is based on #200


floogulinc:
We have cherrypicked this into #221

PR #208, opened 2021-01-17
Integrate the compatibility between bees and flowers into the gameplay

UtkarshKr007:
Closes: #191 

- Students (their bees) can interact with any flower, but only collect pollen from compatible flowers
- After interacting with an incompatible flower:

1.  A tip informs the students about their bee's incompatibility with the flower as well as the fact that no pollen was collected.
2. The interact button for that flower is disabled.

- The tip and the disabling of the interaction button stays until a valid interaction is made.

### DIFF
--- a/src/app/components/play-round/play-round.component.ts
+++ b/src/app/components/play-round/play-round.component.ts
@@ -16,17 +16,20 @@ import { DomSanitizer } from '@angular/platform-browser';
 })
 export class PlayRoundComponent implements OnInit {
   flowerArMarkers$: Observable<RoundMarker[]> = combineLatest([
+    this.studentRoundService.currentBeeSpecies$,
     this.studentRoundService.currentFlowers$,
     this.studentRoundService.currentBeePollen$,
     this.studentRoundService.recentFlowerInteractions$,
   ]).pipe(
-    map(([flowers, beePollen, recentInteractions]) =>
+    filter(([bee]) => !!bee),
+    map(([bee, flowers, beePollen, recentInteractions]) =>
       flowers.map((flower, index) =>
         roundMarkerFromRoundFlower(
           flower,
           index + 1,
           beePollen,
           recentInteractions,
+          !bee.flowers_accepted.map(acceptedFlower => acceptedFlower.id).includes(flower.species.id)
         )
       )
     )
--- a/src/app/components/play-round/play-round.component.ts
+++ b/src/app/components/play-round/play-round.component.ts
@@ -21,7 +21,6 @@ export class PlayRoundComponent implements OnInit {
     this.studentRoundService.currentBeePollen$,
     this.studentRoundService.recentFlowerInteractions$,
   ]).pipe(
-    filter(([bee]) => !!bee),
     map(([bee, flowers, beePollen, recentInteractions]) =>
       flowers.map((flower, index) =>
         roundMarkerFromRoundFlower(
--- a/src/app/components/play-round/play-round.component.ts
+++ b/src/app/components/play-round/play-round.component.ts
@@ -29,7 +29,7 @@ export class PlayRoundComponent implements OnInit {
           index + 1,
           beePollen,
           recentInteractions,
-          !bee.flowers_accepted.map(acceptedFlower => acceptedFlower.id).includes(flower.species.id)
+          bee
         )
       )
     )
--- a/src/app/components/play-round/play-round.component.ts
+++ b/src/app/components/play-round/play-round.component.ts
@@ -111,7 +111,7 @@ export class PlayRoundComponent implements OnInit {
   }
 
   clickInteract(marker: RoundMarker) {
-    this.studentRoundService.interact(marker.barcodeValue, marker.isNest);
+    this.studentRoundService.interact(marker.barcodeValue, marker.isNest, marker.incompatibleFlower ?? false);
   }
 
   calculateBeeScale(scale: number) {
--- a/src/app/markers.ts
+++ b/src/app/markers.ts
@@ -1,3 +1,4 @@
+import { BeeSpecies } from './bees';
 import { RoundFlower, Interaction } from './round';
 
 /**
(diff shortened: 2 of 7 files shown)


src/app/markers.ts
```diff
@@ -53,9 +58,11 @@ export function roundMarkerFromRoundFlower(
   flower: RoundFlower,
   barcodeValue: number,
   currentBeePollen: number,
-  recentFlowerInteractions: Interaction[]
+  recentFlowerInteractions: Interaction[],
+  incompatibleFlower: boolean
 ): RoundMarker {
-  const canVisit = canVisitFlower(
+  const lastVisitedIncompatible = visitingIncompatible(barcodeValue, recentFlowerInteractions);
```
floogulinc, line 64: Since it's only used once, I think we can collapse the function here.

```suggestion
  const lastVisitedIncompatible = recentFlowerInteractions[0].incompatibleFlower && recentFlowerInteractions[0]?.barcodeValue === barcodeValue
```
(and remove the function below)

src/app/markers.ts
```diff
@@ -68,9 +75,16 @@ export function roundMarkerFromRoundFlower(
     isBlooming: flower.isBlooming,
     isNest: false,
     canVisit,
+    incompatibleFlower,
+    tip: lastVisitedIncompatible ? 'I don\'t like this flower. No Pollen collected' : null
```
floogulinc, line 79: I'm not sure we need the "No Pollen collected" part since we will be showing a different animation and the pollen count won't increase.

src/app/markers.ts
```diff
@@ -68,9 +75,16 @@ export function roundMarkerFromRoundFlower(
...
   };
 }
 
+
+export function visitingIncompatible(currentBarcode: number, interactions: Interaction[]) {
+  return((interactions[0]?.barcodeValue === currentBarcode && interactions[0].incompatibleFlower) ?? false);
+}
+
```
floogulinc, line 87:
```suggestion
```

See above suggestion.

src/app/markers.ts
```diff
@@ -85,6 +99,7 @@ export function canVisitFlower(
   recentFlowerInteractions: Interaction[],
 ): boolean {
   const haveVisitedThisFlower = recentFlowerInteractions
+    .filter(interaction => !interaction.incompatibleFlower)
```
floogulinc, line 102: Oh this is interesting, so unlike normal flowers, incompatible flowers aren't affected by the "pollen cycle" lockout.

floogulinc (changes requested):
I wonder if we should have just had a "last interaction" observable in the student round service instead of doing all this filtering everywhere.

src/app/components/play-round/play-round.component.ts
```diff
@@ -16,17 +16,20 @@ import { DomSanitizer } from '@angular/platform-browser';
...
-    map(([flowers, beePollen, recentInteractions]) =>
+    filter(([bee]) => !!bee),
+    map(([bee, flowers, beePollen, recentInteractions]) =>
       flowers.map((flower, index) =>
         roundMarkerFromRoundFlower(
           flower,
           index + 1,
           beePollen,
           recentInteractions,
+          !bee.flowers_accepted.map(acceptedFlower => acceptedFlower.id).includes(flower.species.id)
```
floogulinc, line 32: I think you should pass the bee object into this function and have the function do this part.

src/app/components/play-round/play-round.component.ts
```diff
@@ -16,17 +16,20 @@ import { DomSanitizer } from '@angular/platform-browser';
 })
 export class PlayRoundComponent implements OnInit {
   flowerArMarkers$: Observable<RoundMarker[]> = combineLatest([
+    this.studentRoundService.currentBeeSpecies$,
     this.studentRoundService.currentFlowers$,
     this.studentRoundService.currentBeePollen$,
     this.studentRoundService.recentFlowerInteractions$,
   ]).pipe(
-    map(([flowers, beePollen, recentInteractions]) =>
+    filter(([bee]) => !!bee),
```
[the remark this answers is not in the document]
UtkarshKr007: In the play-round tests, we pass in null for mockBeeSpecies before each test which results in the error where it's trying to do bee.accepted_flower of something that doesn't exists.

src/app/markers.ts
UtkarshKr007: Should we keep it until the animation is added and then that part of the tip can be removed ?

src/app/components/play-round/play-round.component.ts
UtkarshKr007: I am not averse to either way. Is there a benefit in passing primitive type rather than object reference ? I feel like marker.ts doesn't need to know the BeeSpecies type, so I didn't want to import it for a single parameter.

src/app/markers.ts
UtkarshKr007: Yeah. I was wondering why the last interaction logic wasn't working, but the interaction was disabled simply due to incompatible Flower still being a recent interaction.

src/app/components/play-round/play-round.component.ts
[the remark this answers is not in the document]
floogulinc: That seems like almost more of a flaw in the tests but I'd have to check it out more.

src/app/components/play-round/play-round.component.ts
floogulinc: We are already passing in the entire flower, it feels like if we're calling a function to do the work we shouldn't be doing any of the logic here.

UtkarshKr007: Alright, moved the logic to the roundMarkerFromRoundFlower function

src/app/components/play-round/play-round.component.ts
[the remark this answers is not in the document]
UtkarshKr007: I changed the play-round tests to assume the existence of a beeSpecies before each test.

src/app/markers.ts
```diff
@@ -59,9 +60,14 @@ export function roundMarkerFromRoundFlower(
   barcodeValue: number,
   currentBeePollen: number,
   recentFlowerInteractions: Interaction[],
-  incompatibleFlower: boolean
+  bee: BeeSpecies
 ): RoundMarker {
-  const lastVisitedIncompatible = recentFlowerInteractions[0]?.incompatibleFlower && recentFlowerInteractions[0]?.barcodeValue === barcodeValue
+  const incompatibleFlower = !bee.flowers_accepted.map(acceptedFlower =>
+    acceptedFlower.id).includes(flower.species.id);
```
floogulinc, line 66: This is not the best place to split this line. It might be just fine to have this on a single line too.

UtkarshKr007: Doesn't break the TSLint length, but It looks a little long on a single line.

src/app/markers.ts
```diff
@@ -68,6 +80,8 @@ export function roundMarkerFromRoundFlower(
     isBlooming: flower.isBlooming,
     isNest: false,
     canVisit,
+    incompatibleFlower,
+    tip: lastVisitedIncompatible ? 'I don\'t like this flower. No Pollen collected' : null
```
floogulinc, line 84:
```suggestion
    tip: lastVisitedIncompatible ? `${bee.name}s can't collect pollen from this flower` : null
```

KK, Joe, and I discussed some of the phrasing stuff and concluded that we want to avoid the explicit first-person phrasing.

floogulinc (changes requested):

audreylemeur (approved):
It was kinda hard to find an active and non-viable flower (especially with the rusty patch bumblebee) but looks good.

floogulinc (approved):

kklamberty (approved):
This looks good. Thanks for doing this!

PR #213, opened 2021-01-20
Switch to loading our custom icons as constants in the code

floogulinc:
This reduces the number of separate files that need to be requested as these icons are very small and can just be included right in the typescript.


kklamberty (commented):
Can you please help me understand more about why this change is helpful? Thanks!

floogulinc:
> Can you please help me understand more about why this change is helpful? Thanks!

It mainly reduces the number of small requests required to load certain pages by not loading the tiny SVG icons from separate requests and just including them right in the JS bundle. This is something I didn't know was possible otherwise I think we would have done it this way from the beginning.

UtkarshKr007 (approved):
All the icons still load for me

kklamberty (approved):
That makes sense. Thanks!

PR #216, opened 2021-01-25
Get the GitHub SVG from the right place

### LINKED ISSUE #215: The GitHub icon isn't appearing on the about page
Copy-paste of the description from [here](https://github.com/mn-pollinators/buzz-about/pull/206#issuecomment-766485276):

> Yep, okay, it looks like these lines in `about.component.ts` are the problem:
> 
> ```ts
>   constructor(iconRegistry: MatIconRegistry, sanitizer: DomSanitizer, public updates: SwUpdate) {
>     iconRegistry.addSvgIcon(
>       'github',
>       sanitizer.bypassSecurityTrustResourceUrl('assets/github.svg')
>     );
>   }
> ```
> 
> The constructor is still trying to load the SVG icon from assets, even though it doesn't live there anymore. (It's actually shadowing the correct version of the icon with the broken version.)
> 
> I'll make a quick PR to fix that.
> 
> Sorry—my fault for re-approving the changes without looking at the deployment :(

helloworld12321:
We recently changed where the GitHub icon SVG lives. It used to be in the `assets/` directory; now it's in a TS file. (We're statically linking it in with the WebPack bundle.)

Unfortunately, one lingering reference to the `assets/` directory didn't get changed.  (Or, it got changed, and then it got changed back, or something like that.) This commit fixes the issue.

(I also quickly searched through the codebase, and it looks like that was the last reference to storing custom SVG icons in `assets/`.)

Closes  #215.

### DIFF
--- a/src/app/pages/about/about.component.ts
+++ b/src/app/pages/about/about.component.ts
@@ -1,5 +1,4 @@
 import { Component, OnInit } from '@angular/core';
-import { MatIconRegistry } from '@angular/material/icon';
 import { DomSanitizer } from '@angular/platform-browser';
 import { SwUpdate } from '@angular/service-worker';
 import { buzzAbout as buzzAboutInfo, assets as assetsInfo } from '../../../../project-info.json';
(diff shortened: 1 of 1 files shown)


floogulinc:
Good catch Joe, must have been a bad merge or something.

floogulinc (approved):

kklamberty (approved):

kklamberty:
Thanks for fixing that and checking for other instances of similar issues!

PR #221, opened 2021-02-02
Add a thought bubble next to the bee to indicate various states

### LINKED ISSUE #218: Add bee thought bubble to indicate different hint states
We want to have a thought bubble to the left of the bee that can show different icons for different states, for example:
- a question mark when the marker you're looking at is one you tried and failed to interact with
- a nest icon or image when you are full on pollen
- a flower icon when at your nest and empty of pollen

We found a thought bubble SVG to try that we liked: https://publicdomainvectors.org/en/free-clipart/Vector-clip-art-of-thick-border-thought-cloud/21095.html

floogulinc:
Closes #218 

| Full of Pollen | Incompatible Flower |
|---|---|
| image | image |


floogulinc:
We have decided to remove the "at nest but have no pollen" thought bubble.

src/app/components/play-round/play-round.component.ts
```diff
@@ -158,6 +156,8 @@ export class PlayRoundComponent implements OnInit {
   }
 
   async clickInteract() {
+    this.interactionInProgress = true;
```
helloworld12321, line 159: It might be technically possible for a use to click the interact button, and then hit it again before this assignment finishes,

but I think that for practical purposes, setting this flag should work all right ☺️

Thanks for setting that up, Paul!

helloworld12321: (Please excuse me, I'm just always a bit paranoid about async code 😛)

helloworld12321 (commented):
I took a look at the async stuff you added—thanks for doing all that! :)

I'm not actually going to approve this pull request, on account of the fact that I worked on it, but thanks for bringing me over to come take a look at all the new stuff—it's pretty cool! 😃

src/app/components/play-round/play-round.component.ts
floogulinc: I don't think there's really any faster way to disable the button upon clicking it. I suppose we could add some sort of debounce in addition to this but in my testing it seems extremely unlikely to be an issue.

kklamberty (changes requested):
I'm not sure what is happening to cause this, but when I try to create a session join code, it is not working. I've never run into that before today. When I try to "rejoin" a session, a join code does show up, so it's not that it isn't being created (I don't think). 

Do you think something with the asynchronous stuff is causing trouble? 

This is also the first time I'm having major weirdness between the phone (waiting for a round to start that has definitely been started... when I paused it in late September, my phone still didn't know the round had started, but it eventually did say "Paused"). I think something is quite different than it used to be.

audreylemeur (approved):

kklamberty (approved):

PR #227, opened 2021-02-16
Add some new rounds for production use

helloworld12321:
Previously, we had some demo round-templates for internal testing, but no round-templates that we really wanted the students to play.

This pull request removes the demo rounds, and adds three round templates we're actually planning on using in production (although we'll probably make edits and revisions, of course).

They're designed so that students will learn that different bees need different flowers to survive; and that, if you only plant one or two types of flowers, you're only serving the needs of a few bee species.

### DIFF
--- a/src/app/round-template.ts
+++ b/src/app/round-template.ts
@@ -43,6 +43,29 @@ export const roundTemplates: RoundTemplate[] = [
     ]
   },
   {
+    name: 'Field #1: Four Flower Species',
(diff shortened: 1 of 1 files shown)


floogulinc (approved):

floogulinc (approved):

UtkarshKr007 (approved):

helloworld12321:
Thanks very much—merging!

PR #230, opened 2021-02-16
Add round sets and other round choosing improvements 

floogulinc:
In the round-chooser dialogue, we're grouping rounds into sets (that is, progressions of rounds with a common learning goal).

Also, we added a "description" field for both rounds and sets. These descriptions are displayed when you're choosing a round.

This PR depends on PR #227 "Add some new rounds for production use". 

image


UtkarshKr007 (dismissed):
Great work fixing the bug and on the PR in genereal.

audreylemeur:
It doesn't look great on my smaller laptop screen (14''). Do we think that would be a problem in general?
image

floogulinc:
@audreylemeur What is the resolution and scaling factor of your display? Namely these two numbers: 
image

PR #234, opened 2021-02-24
Remove customAssign function and change templates to just take an array of bee species

### LINKED ISSUE #229: Round templates should not need weights for bees
Currently, if you give a list of bees, then you *must* give weights for those bees. By default, if you give a list of bees, perhaps it should just make all of them have equal weights without making you enter any weights at all.

floogulinc:
We are moving to the templates taking an array of bee species rather than an array of bees with weights. If you want to have some bee species weighted differently you can duplicate that species in the array. This allows us to use the much simpler and more reliable assignment algorithm.

This PR is based on #230 so it should only be merged after #230 is.

Closes #229


src/app/services/teacher-round.service.ts
```diff
@@ -92,17 +93,27 @@ export class TeacherRoundService {
...
-    // Get the list of students in the session
-    if (bees) {
-      return this.customAssign(studentList, bees, currentRoundPath);
-    } else {
-      return this.defaultAssign(studentList, currentRoundPath);
-    }
+    const shuffledBees = shuffleArray<BeeSpecies>(bees);
+
+    await Promise.all(studentList.map((student, studentIndex) => {
+      const beeIndex = studentIndex % shuffledBees.length;
```
helloworld12321, line 112: Okay, that's pretty slick.

helloworld12321 (approved):
Looks awesome. (And it works well, too!)
Thanks, Paul! :)

audreylemeur (approved):

PR #237, opened 2021-02-26
Refactor fullscreen button to use NgxScreenfull, add back to bottom bar

floogulinc:
Removes our old fullscreen button logic and replaces it with the [@ngx-extensions/screenful](https://www.npmjs.com/package/@ngx-extensions/screenfull) library. Also adds the fullscreen button back to the bottom bar replacing the menu button that wasn't being used.


UtkarshKr007 (dismissed):
I really like the big display in full screen mode. Great addition. We also provide a cancel button or a back button on every page that can be full screen, so the loss of browser's back button doesn't affect any sort of navigation.

floogulinc:
> I really like the big display in full screen mode. Great addition. We also provide a cancel button or a back button on every page that can be full screen, so the loss of browser's back button doesn't affect any sort of navigation.

I made it so quitting the session leaves fullscreen so we really only intend the session lobby and onward to be fullscreen and don't have to worry about providing a fullscreen button anywhere else.

floogulinc:
Looks like there is a bug when you start a round while being fullscreen the button icon reset.

UtkarshKr007 (approved):
The full screen icon reseting bug is fixed and everything else looks great just as before.

kklamberty (approved):
This looks great to me! I tried out a bunch of different things to try to break it and it seems pretty robust. Nice work!

PR #238, opened 2021-02-26
Update GitHub Actions badges in README.md to new format

floogulinc:


helloworld12321 (approved):
Looks good, thanks, Paul! :)

kklamberty (approved):
These don't look different, but they do seem to be working. I'm guessing the new format has to do with how the work is requested and not how the badges look. The newer version certainly looks more straightforward.

Issue #243, opened 2021-03-10
Add a 404 page


helloworld12321:
Right now we're just showing a blank screen, which isn't very intuitive. We probably want to have a page that more explicitly says "Not found" or something like that.

<img width="612" alt="Screen Shot 2021-03-10 at 3 54 24 PM" src="https://user-images.githubusercontent.com/56209343/110703054-e91eaf80-81b8-11eb-9ed3-61fe865dd752.png">

<div align="center">⬇️⬇️⬇️</div align="center">

<img width="612" alt="Screen Shot 2021-03-10 at 3 52 14 PM" src="https://user-images.githubusercontent.com/56209343/110702786-96dd8e80-81b8-11eb-824b-402eb29c5995.png">

PR #244, opened 2021-03-10
Create 404 page

UtkarshKr007:
Add a 404 page for any non-defined routes.

Closes: #243


kklamberty (dismissed):
Looks good to me!

src/app/pages/page-not-found/page-not-found.component.html
```diff
@@ -0,0 +1,13 @@
...
+<div class="centered-container">
+  <div>
+    <mat-card>
+      <mat-card-title class="title">Whoops</mat-card-title>
+      <mat-card-content>
+        This page doesn't seem to exist.
+      </mat-card-content>
+    </mat-card>
+  </div>
+</div>
```
floogulinc, line 12: I think we should try to have all our errors be consistent so this would change it to be more like the invalid round error and others:
```suggestion
<app-hill-background cover></app-hill-background>
<div class="centered-container">
  <div>
    <mat-card>
      <mat-card-title>Whoops</mat-card-title>
      <mat-card-content>
        This page doesn't seem to exist.
      </mat-card-content>
      <mat-card-actions align="end">
        <button mat-button color="primary" routerLink="/">Go Back</button>
      </mat-card-actions>
    </mat-card>
  </div>
</div>
```

We might want to say "Go Home" rather than "Go Back" but I'm not completely sure.

src/app/pages/page-not-found/page-not-found.component.html
```diff
@@ -0,0 +1,13 @@
+<app-hill-background cover></app-hill-background>
+<app-back-button top-left routerLink="/"></app-back-button>
+<div class="centered-container">
+  <div>
+    <mat-card>
+      <mat-card-title class="title">Whoops</mat-card-title>
```
floogulinc, line 6:
```suggestion
      <mat-card-title>Page Not Found</mat-card-title>
```

I prefer more informative wording here.

src/app/pages/page-not-found/page-not-found.component.html
```diff
@@ -0,0 +1,13 @@
...
+      <mat-card-content>
+        This page doesn't seem to exist.
```
floogulinc, line 8: I think we should mention the route that's being attempted here. People using the app added to their home screen won't be able to see what route caused the error otherwise.

src/app/pages/page-not-found/page-not-found.component.scss
```diff
@@ -0,0 +1,3 @@
+.title {
+  text-align: center;
+}
```
floogulinc, line 3:
```suggestion
```

We don't center card titles on other error cards/dialogs. (Though, if you really want to you should use our `centered-text` class.)

src/app/pages/page-not-found/page-not-found.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'app-page-not-found',
+  templateUrl: './page-not-found.component.html',
+  styleUrls: ['./page-not-found.component.scss']
```
floogulinc, line 6:
```suggestion
  styleUrls: ['./page-not-found.component.scss'],
  changeDetection: ChangeDetectionStrategy.OnPush
```

floogulinc (changes requested):
I have made a few suggestions mainly around having this look and feel like our other error messages/dialogs.

src/app/pages/page-not-found/page-not-found.component.ts
```diff
@@ -0,0 +1,20 @@
...
+  selector: 'app-page-not-found',
+  templateUrl: './page-not-found.component.html',
+  styleUrls: ['./page-not-found.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class PageNotFoundComponent implements OnInit {
+  path: string
+
+  constructor(private route: ActivatedRoute) {
+    this.path = "\"/" + this.route.snapshot.url.join('/') + "\"";
```
floogulinc, line 14:
```suggestion
constructor(public route: ActivatedRoute) {
```

I think we can just use the observable from `ActivatedRoute` right in the template (see other suggestion).

src/app/pages/page-not-found/page-not-found.component.html
```diff
@@ -0,0 +1,14 @@
+<app-hill-background cover></app-hill-background>
+<div class="centered-container">
+  <div>
+    <mat-card>
+      <mat-card-title>Page Not Found</mat-card-title>
+      <mat-card-content>
+        Sorry, we couldn't find a page located at {{path}}.
```
floogulinc, line 7:
```suggestion
        Sorry, we couldn't find a page at <code>{{(route.url | async).join('/')}}</code>.
```

floogulinc (changes requested):

audreylemeur (commented):
Just wanted to let you know that I've looked at this pull request. Visually and functionally it looks good so once the test situation gets figured out, I think it can be approved.

src/app/pages/page-not-found/page-not-found.component.spec.ts
```diff
@@ -1,8 +1,7 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
-
 import { PageNotFoundComponent } from './page-not-found.component';
 
-describe('PageNotFoundComponent', () => {
+xdescribe('PageNotFoundComponent', () => {
```
floogulinc, line 4: I think we should have the tests enabled here even if it's just the "should create" test.

UtkarshKr007: That would mean we mock activated route which I am fine with if we add more test, but I don't think its worth it for just a "should exit" default test.

floogulinc: Yeah, that makes sense, I guess we are just disabling tests on many of the other pages as well.

src/app/pages/page-not-found/page-not-found.component.html
```diff
@@ -0,0 +1,14 @@
+<app-hill-background cover></app-hill-background>
+<div class="centered-container">
+  <div>
+    <mat-card>
+      <mat-card-title>Page Not Found</mat-card-title>
+      <mat-card-content>
+        Sorry, we couldn't find a page at <code>/{{(route.url | async).join('/')}}</code>.
+      </mat-card-content>
+      <mat-card-actions align="end">
+        <button mat-button color="primary" routerLink="/">Navigate to Home</button>
```
floogulinc, line 10:
```suggestion
        <button mat-button color="primary" routerLink="/">Go Home</button>
```

floogulinc (changes requested):

floogulinc (approved):

audreylemeur (approved):

PR #256, opened 2021-04-06
Add the ability for teachers to display the field guide for students

floogulinc:
image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -98,5 +98,6 @@
   "cSpell.enableFiletypes": [
     "firerules"
   ],
-  "markdown.extension.toc.levels": "2..6"
+  "markdown.extension.toc.levels": "2..6",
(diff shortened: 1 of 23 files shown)


kklamberty (changes requested):
I'm having some kind of issue with the field guide not always showing the images when I'm looking at an individual item. I'm not sure why. When I refresh, the image will show up. If I'm looking at an individual item, it's also the case that if the host chooses to stop showing the field guide, I can still see the element I'm looking at.

Is there anything I can do to help someone recreate the issue about not being able to see images?

floogulinc:
@kklamberty We have fixed the issue of the dialog staying open. We have yet to look at the image loading issue but suspect it may just be an issue with Safari that we may not really be able to fix.

kklamberty (approved):

audreylemeur (approved):

PR #259, opened 2021-04-15
Add fix for field guide images in Webkit

floogulinc:
This is a (hopefully temporary) fix for the issue we were having where images wouldn't display in the field guide dialog in Webkit (Safari). We now replace the images in the field guide dialog with the 512 square versions only on Webkit so we can set a default width/height for those.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2846,7 +2846,8 @@
         },
         "node-fetch": {
           "version": "2.6.0",
-          "resolved": "",
+          "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.6.0.tgz",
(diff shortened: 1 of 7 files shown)


kklamberty (approved):
I think this fixes the problem; I always had an image. On my iPhone, I noticed that when I close the dialog about an individual field guide element there is a flash like things are loading twice. This doesn't happen in Safari on my desktop, and it doesn't see to happen with the elements that show before I scroll (those first 4 flowers). Does that relate directly to part of the fix or is it something to try to capture and look into?

audreylemeur (commented):
The photos all load but they look very squished on my iPhone SE.
image

floogulinc:
> The photos all load but they look very squished on my iPhone SE.

@audreylemeur This only happens on iOS 10, which we don't support.

helloworld12321 (approved):
(Approving, even though I worked on this PR, because not many people in this project have devices on which to test this fix.)

PR #262, opened 2021-04-20
Optimize and cleanup AR view

floogulinc:
Cleans up and optimizes a bunch of code in the AR view component including using async await for the initAR function and reverting to doing the disposal stuff right away when ngOnDestroy is called. 

Also adds throttling to the teacher round service running status output to Firebase.


floogulinc:
Superseded by #269

Issue #268, opened 2021-04-27
Rotation on iOS messes up the size and positioning of the THREEAR canvas element.


helloworld12321:
This issue is unrelated to #263. But you can see it happening in the same video:
(Note how the THREEAR canvas is misaligned, so that you can see white borders on the sides of the screen.)

https://user-images.githubusercontent.com/56209343/116304557-4de19980-a768-11eb-94e8-0a3ab587a29e.mov

The problem is that on iOS Safari, when you rotate the screen, there's a brief "rotation" animation, and during that animation, (`window.innerWidth`)[https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth] and `window.innerHeight` have nonsensical values.

See this GitHub issue: https://stackoverflow.com/questions/4629969/ios-return-bad-value-for-window-innerheight-width

This issue occurs both when Buzz About is run in the browser and when it's added to the home screen.

Issue #263, opened 2021-04-22
Rotation on iOS zooms in (irrecoverably)


helloworld12321:
We've disabled pinch-to-zoom, so you can't zoom back out 😢

https://user-images.githubusercontent.com/56209343/115767896-3a50c000-a36f-11eb-9685-3f55ef56bca7.mov

PR #269, opened 2021-04-27
AR bug fixes and viewport locking

floogulinc:
Cleans up and optimizes a bunch of code in the AR view component including using async await for the initAR function and reverting to doing the disposal stuff right away when ngOnDestroy is called.

Also adds throttling to the teacher round service running status output to Firebase.

This PR cleans up and optimizes a bunch of code in the AR view component including using async await for the initAR function and reverting to doing the disposal stuff right away when ngOnDestroy is called.

This also sets the viewport scale to be locked at 1 (not so great for accessibility but it's similar to how native apps works and really helps with the AR zooming issues on iOS). 

Specifics about locking the viewport scale:
- iOS Safari when added to homescreen will not allow any pinch to zoom or automatic zooming. The AR view works pretty much perfectly now in this mode.
- iOS Safari in the regular browser will still allow pinch to zoom but no automatic zooming so it will not automatically zoom in on the join form and will not automatically zoom when rotated (#263) but users can still pinch to zoom which can still break the AR view.
- Most other browsers including Chrome on Android will not allow pinch to zoom in any situation.

We also switched to our [own fork of THREEAR](https://github.com/mn-pollinators/THREEAR) to fix some more issues in Safari when rotating the screen (#268).

Also adds throttling to the teacher round service running status output to Firebase.


UtkarshKr007 (approved):
Tried it on android phone. Really like how the native app is full screen now. Feels much more immersive. I went through several rounds without an issue with anything else.

kklamberty (approved):
The process of testing for reviewing this pull request revealed a minor issue with stretching the images of flowers and nests, but this PR does fix several other things and should be merged without waiting for a fix on that other issue.

PR #271, opened 2021-04-29
Show bees on large display flower field

floogulinc:
https://buzz-about-git-large-display-bees-mn-pollinators.vercel.app/test/flower-test


kklamberty (approved):
I love how the animation looks! Thanks for your work on this. I've noticed that when I create a session and then create a join code, I sometimes need to refresh the page to make my join code appear. That is not relevant to this PR, but may need some sleuthing at some point.

PR #278, opened 2021-05-19
Pre and post round screens

floogulinc:


floogulinc:
Ideas from meeting:
- show average pollen totals from each flower on large display

floogulinc:
- [ ] check what happens with fractional number average pollen, (rounding or truncation?)
- [ ] show total and per species number of bees

PR #288, opened 2021-06-29
Add 7 min demo video and IDC 2021 paper to README

floogulinc:


kklamberty (approved):
The links are all correct URLs and these changes look good to me. Thanks for doing this!

